PR #2, opened 2018-07-23
Added the barebones structure skeleton and barebones pom.xml

drosser:
The pom.xml will be updated with dependencies as and when the new dependencies are needed (e.g. surefire plugin, sonar dependency).

Resolves: SFA-43.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,11 +4,11 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>uk.gov.companieshouse</groupId>
-    <artifactId>web.accounts</artifactId>
+    <artifactId>company-accounts.web.ch.gov.uk</artifactId>
     <version>0.1.0</version>
     <packaging>jar</packaging>
 
-    <name>web.accounts</name>
+    <name>company-accounts-web</name>
     <description>Web service to handle company account data</description>
 
     <parent>
--- a/pom.xml
+++ b/pom.xml
@@ -40,12 +40,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.thymeleaf.extras</groupId>
-            <artifactId>thymeleaf-extras-java8time</artifactId>
-            <version>${thymeleaf-extras.version}</version>
-        </dependency>
-
-        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>${lombok.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>web.accounts</artifactId>
+    <version>0.1.0</version>
+    <packaging>jar</packaging>
+
+    <name>web.accounts</name>
+    <description>Web service to handle company account data</description>
+
+    <parent>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-parent</artifactId>
+        <version>2.0.3.RELEASE</version>
+        <relativePath/> <!-- lookup parent from repository -->
+    </parent>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <java.version>1.8</java.version>
+
+        <lombok.version>1.18.0</lombok.version>
+        <thymeleaf-extras.version>3.0.1.RELEASE</thymeleaf-extras.version>
+        <structured-logging.version>1.2.0-rc1</structured-logging.version>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-thymeleaf</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.thymeleaf.extras</groupId>
+            <artifactId>thymeleaf-extras-java8time</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -0,0 +1,90 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>web.accounts</artifactId>
+    <version>0.1.0</version>
+    <packaging>jar</packaging>
+
+    <name>web.accounts</name>
```
cburns93: Can we follow the same naming as the company-api?
` <name>company-accounts-api</name>` So` <name>company-accounts-web</name>`

pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>web.accounts</artifactId>
+    <version>0.1.0</version>
```
jsimons1989: This shouldn't have a hardcoded release version

pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    </parent>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <java.version>1.8</java.version>
+
+        <lombok.version>1.18.0</lombok.version>
+        <thymeleaf-extras.version>3.0.1.RELEASE</thymeleaf-extras.version>
+        <structured-logging.version>1.2.0-rc1</structured-logging.version>
```
cburns93: Not the latest version of structured logging?

pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>web.accounts</artifactId>
```
jsimons1989: We have followed a standard of the artifactId matching the repository name.

pom.xml
```diff
@@ -0,0 +1,90 @@
...
+                        <goals>
+                            <goal>repackage</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+
```
cburns93: I don't see any resolution repositories? How is someone supposed to build the project if you haven't specified the artifactory in which the dependencies are located?

src/main/java/uk/gov/companieshouse/web/accounts/Application.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.companieshouse.web.accounts;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class Application {
```
cburns93: can we rename this to follow company accounts api? to `CompanyAccountsApplication`? Just to keep consistent

pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.thymeleaf.extras</groupId>
+            <artifactId>thymeleaf-extras-java8time</artifactId>
+            <version>${thymeleaf-extras.version}</version>
```
cburns93: It is brought in with the`spring-boot-starter-thymeleaf` so don't need this depedency. Additionally it is always best to let the parent bring the dependencies in instead of declaring them manually as you may end up forgetting to update the version which then becomes out of sync. https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-thymeleaf/2.0.3.RELEASE

pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>uk.gov.companieshouse</groupId>
+            <artifactId>structured-logging</artifactId>
+            <version>${structured-logging.version}</version>
```
cburns93: Not using latest version of structured logging?

pom.xml
cburns93: Agreed. It should be `unversioned`

pom.xml
```diff
@@ -0,0 +1,90 @@
...
+
+        <lombok.version>1.18.0</lombok.version>
+        <thymeleaf-extras.version>3.0.1.RELEASE</thymeleaf-extras.version>
+        <structured-logging.version>1.2.0-rc1</structured-logging.version>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
```
cburns93: This is brought in with `spring-boot-starter-parent`.

pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <artifactId>spring-boot-starter-thymeleaf</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
```
cburns93: This is brought in with `spring-boot-starter-parent`.

pom.xml
drosser: This is the barebones pom, and doesn't contain everything needed. The repository also doesn't contain all the needed dependencies for the repositories, doesn't contain a `Makefile`, nor a start script.

pom.xml
drosser: Need to do some more fiddling with dependency management for this to be true. Will leave it as-is for now.

pom.xml
drosser: As above.

pom.xml
cburns93: My comment referenced the artifactories because you had structured logging as a dependency but as it has been removed my first comment is redundant as all dependencies are at mvncentral.

pom.xml
cburns93: 👍

PR #3, opened 2018-07-31
 Add Makefile and start script to allow for service to be built and started

drosser:
Resolves: SFA-43.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,46 @@
+artifact_name       := company-accounts.web.ch.gov.uk
+commit              := $(shell git rev-parse --short HEAD)
+tag                 := $(shell git tag -l 'v*-rc*' --points-at HEAD)
+version             := $(shell if [[ -n "$(tag)" ]]; then echo $(tag) | sed 's/^v//'; else echo $(commit); fi)
+artifactory_publish := $(shell if [[ -n "$(tag)" ]]; then echo release; else echo dev; fi)
+
+.PHONY: all
+all: build
+
+.PHONY: clean
+clean:
+	mvn clean
+	rm -f ./$(artifact_name).jar
+	rm -f ./$(artifact_name)-*.zip
+	rm -rf ./build-*
+	rm -f ./build.log
+
+.PHONY: build
+build:
+	mvn versions:set -DnewVersion=$(version) -DgenerateBackupPoms=false
+	mvn package -DskipTests=true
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,46 @@
+artifact_name       := company-accounts.web.ch.gov.uk
+commit              := $(shell git rev-parse --short HEAD)
+tag                 := $(shell git tag -l 'v*-rc*' --points-at HEAD)
+version             := $(shell if [[ -n "$(tag)" ]]; then echo $(tag) | sed 's/^v//'; else echo $(commit); fi)
+artifactory_publish := $(shell if [[ -n "$(tag)" ]]; then echo release; else echo dev; fi)
```
cburns93: `artifactory_publish` isn't used anywhere? If you haven't made a `publish` target for a reason, then remove this line out.

start.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo "Downloading environment from: ${CONFIG_URL}/${ENVIRONMENT}/${APP_NAME}"
+    wget -O "${APP_DIR}/private_env" "${CONFIG_URL}/${ENVIRONMENT}/private_env"
+    wget -O "${APP_DIR}/global_env" "${CONFIG_URL}/${ENVIRONMENT}/global_env"
+    wget -O "${APP_DIR}/app_env" "${CONFIG_URL}/${ENVIRONMENT}/${APP_NAME}/env"
+    source "${APP_DIR}/private_env"
+    source "${APP_DIR}/global_env"
+    source "${APP_DIR}/app_env"
+fi
+
+exec java ${JAVA_MEM_ARGS} -jar -Dserver.port="${PORT}" "${APP_DIR}/company-accounts.web.ch.gov.uk.jar"
```
cburns93: It may be worth adding the debug settings within the start script like we have done in the `company-accounts.api.ch.gov.uk`

drosser: We're not going to add a debug command that is commented out. It deviates from our other `start` scripts, but if we did want to support this as standard then we'd add proper functionality with flags (or some other variation) as opposed to adding a commented out debug line.

cburns93: It is definitely worth adding or even possibly discussing. `./start.sh -d` to trigger the debug profile would be handy and additionally making it default to non-debug without it.

drosser: I think this is worth a bigger discussion (tech meeting?), but I don't think this should block a PR going in to add *basic* start script functionality so that this service can be run (bear in mind there's no endpoints in the service at this point).

PR #5, opened 2018-08-01
Add POJO view models for Called Up Share Capital Not Paid (current period only)

drosser:
- Add BalanceSheet model
- Add CalledUpShareCapitalNotPaid model
- Add ValidationMessages.properties

Resolves: SFA-43.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CalledUpShareCapitalNotPaid.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CalledUpShareCapitalNotPaid.java
@@ -1,0 +1,16 @@
+package uk.gov.companieshouse.web.accounts.model.smallfull;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.hibernate.validator.constraints.Range;
+
+import javax.validation.constraints.NotNull;
+
+@Getter
+@Setter
+public class CalledUpShareCapitalNotPaid {
+
+    @NotNull
+    @Range(min = 0, max = 99999999, message = "{invalidRange}")
+    private Integer currentAmount;
+}
--- a/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CalledUpShareCapitalNotPaid.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CalledUpShareCapitalNotPaid.java
@@ -4,13 +4,10 @@ import lombok.Getter;
 import lombok.Setter;
 import org.hibernate.validator.constraints.Range;
 
-import javax.validation.constraints.NotNull;
-
 @Getter
 @Setter
 public class CalledUpShareCapitalNotPaid {
 
-    @NotNull
     @Range(min = 0, max = 99999999, message = "{invalidRange}")
     private Integer currentAmount;
 }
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CalledUpShareCapitalNotPaid.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+import org.hibernate.validator.constraints.Range;
+
+@Getter
+@Setter
+public class CalledUpShareCapitalNotPaid {
+
+    @Range(min = 0, max = 99999999, message = "{invalidRange}")
```
cburns93: I'm just curious, is there a reason why we are using hibernate range instead of the Javax validation?
`import javax.validation.constraints.Size;`
`@Size(min = 10, max = 200, message= "About Me must be between 10 and 200 characters")` I ask in a regards of possibly reducing the amount of external dependencies?

BSaunders95: According to the [documentation](https://docs.oracle.com/javaee/7/api/javax/validation/constraints/Size.html), the `@Size` annotation supports:
- CharSequence (length of character sequence is evaluated)
- Collection (collection size is evaluated)
- Map (map size is evaluated)
- Array (array length is evaluated)

cburns93: It does support multiple types, however it is dependent on the type itself which in this case is an `Integer`. Additionally, how does `@Range` handle nulls? My only concern is that hibernate is an ORM mapper that maps an object-oriented domain model to a relational database. Since we are not using relational databases I just worry that we lead developers the wrong way by using Hibernate without having a RDB in the back end =/

drosser: This dependency isn't an additional dependency to anything we're bringing in currently (i.e. we're not adding an explicit dependency to use this annotation), so we don't have to worry about that aspect.

Regarding `@Range` handling nulls, it doesn't, neither does it need to. If we wanted to stop nulls being entered, we'd just use the `@NotNull` annotation. 

Hibernate is an ORM mapper, but looking at what the annotation itself does, there's no DB-specific logic in there. If we use the `@Size` annotation, we cannot use the `Integer` type against it, and the `@Size` checks the *length* of the sequence (whether that be the number of characters in the String, the size of the map, etc) rather than whether the provided value is within an upper or lower bound.

`@Range` gives us the added advantage of re-using the same message for different parameters - for example, in this usage we're checking the range is between `0` and `99999999`, but it also gives us the consistency that for another field we use the same annotation but the range is between `-1` and `99999999`. Using the `@Size` annotation, this becomes more fiddly. 

I think `@Range` is the best tool for the job, especially with the consistency it give us.

cburns93: I'm pretty sure that you can use `@Size` with Integers, but if all the other points are true, then I'm happy to approve.

PR #6, opened 2018-08-01
Add basic interface to BalanceSheetServiceImpl and BalanceSheetTransformer.

drosser:
- Add basic interface to BalanceSheetServiceImpl that controller will use
- Add basic interface to BalanceSheetTransformer that service will use
- Update pom to import JUnit Jupiter dependency and Mockito dependency
- Add basic unit tests for implementations of service and transformer

Resolves: SFA-43.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
@@ -21,13 +21,4 @@ public class BalanceSheetServiceImpl implements BalanceSheetService {
 
         return balanceSheet;
     }
-
-    /*private CurrentPeriod getCurrentPeriodFromAPI(String companyId, String transactionId, String accountsId) {
-
-
-    }*/
-
-    /*private PreviousPeriod getPreviousPeriodFromAPI(String companyId, String transactionId, String accountsId) {
-
-    }*/
 }
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
@@ -1,6 +1,33 @@
 package uk.gov.companieshouse.web.accounts.service.smallfull.impl;
 
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
 import uk.gov.companieshouse.web.accounts.service.smallfull.BalanceSheetService;
+import uk.gov.companieshouse.web.accounts.transformer.smallfull.BalanceSheetTransformer;
 
+@Component
 public class BalanceSheetServiceImpl implements BalanceSheetService {
+
+    private static BalanceSheetTransformer transformer = new BalanceSheetTransformer();
+
+    @Override
+    public BalanceSheet getBalanceSheet(String companyId, String transactionId, String accountsId) {
+
+        //get data from API
+
+
+        //transform API data to web-readable data
+        BalanceSheet balanceSheet = transformer.getBalanceSheet();
+
+        return balanceSheet;
+    }
+
+    /*private CurrentPeriod getCurrentPeriodFromAPI(String companyId, String transactionId, String accountsId) {
+
+
+    }*/
+
+    /*private PreviousPeriod getPreviousPeriodFromAPI(String companyId, String transactionId, String accountsId) {
+
+    }*/
 }
--- a/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
@@ -1,4 +1,15 @@
 package uk.gov.companieshouse.web.accounts.transformer.smallfull;
 
+import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
+import uk.gov.companieshouse.web.accounts.model.smallfull.CalledUpShareCapitalNotPaid;
+
 public class BalanceSheetTransformer {
+
+    public BalanceSheet getBalanceSheet() {
+
+        BalanceSheet balanceSheet = new BalanceSheet();
+        balanceSheet.setCalledUpShareCapitalNotPaid(new CalledUpShareCapitalNotPaid());
+
+        return balanceSheet;
+    }
 }
--- a/src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
+++ b/src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
@@ -14,6 +14,7 @@ public class BalanceSheetServiceImplTests {
 
     @BeforeEach
     private void init() {
+
         balanceSheetService = new BalanceSheetServiceImpl();
     }
 
--- a/src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
+++ b/src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
@@ -19,7 +19,7 @@ public class BalanceSheetServiceImplTests {
     }
 
     @Test
-    public void getBalanceSheetSuccess() {
+    void getBalanceSheetSuccess() {
 
         BalanceSheet balanceSheet = balanceSheetService.getBalanceSheet(any(String.class), any(String.class), any(String.class));
         assertNotNull(balanceSheet);
--- a/src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
+++ b/src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
@@ -1,0 +1,26 @@
+package uk.gov.companieshouse.web.accounts.service.smallfull.impl;
+
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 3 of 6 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -1,6 +1,24 @@
...
 
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
 import uk.gov.companieshouse.web.accounts.service.smallfull.BalanceSheetService;
+import uk.gov.companieshouse.web.accounts.transformer.smallfull.BalanceSheetTransformer;
 
+@Component
 public class BalanceSheetServiceImpl implements BalanceSheetService {
+
+    private static BalanceSheetTransformer transformer = new BalanceSheetTransformer();
```
cburns93: Is there a reason we are adding the transformer as a static field? and instead not autowiring it into the class?

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
```diff
@@ -1,4 +1,15 @@
 package uk.gov.companieshouse.web.accounts.transformer.smallfull;
 
+import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
+import uk.gov.companieshouse.web.accounts.model.smallfull.CalledUpShareCapitalNotPaid;
+
 public class BalanceSheetTransformer {
+
+    public BalanceSheet getBalanceSheet() {
```
cburns93: Is this method set in stone? Or is it purely just to get something in? I only ask because in the api we have overloaded transform methods that you can call to transform data between objects.

src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class BalanceSheetServiceImplTests {
+
+    BalanceSheetService balanceSheetService;
+
+    @BeforeEach
+    private void init() {
+        balanceSheetService = new BalanceSheetServiceImpl();
```
cburns93: empty line needs to be added for consistency or either remove it elsewhere.

src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    @Test
+    public void getBalanceSheetSuccess() {
```
cburns93: I believe you can get away with removing the access modifier on the tests now :)

src/test/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformerTests.java
```diff
@@ -0,0 +1,27 @@
...
+    private BalanceSheetTransformer transformer;
+
+    @BeforeEach
+    public void init() {
+
+        transformer = new BalanceSheetTransformer();
+    }
+
+    @Test
+    public void getBalanceSheetSuccess() {
```
cburns93: I believe you can get away with removing the access modifier on the tests now :)

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
BSaunders95: We may rename this method, though the problem we're facing is that when converting data to send to the API we'll be taking the balance sheet and converting it to a current period and to a previous period. The method names must be clear and concise to avoid confusion, though the given method name is subject to change.

cburns93: Are you not using any mapping frameworks as we discussed before? Additionally, will you be converting objects from Web to API and vice versa from the top level parent? If so, you should be able to just use bean copy utils

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
BSaunders95: We want it to be static due to the nature of the Transformer (it only needs to be initialised once as nothing differs per Service). We could qualify it with `final` too to help indicate this.

The counter-argument is that we program to an interface, use `@Autowired`, and don't make it static.

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
BSaunders95: We may well end up using `BeanUtils#copyProperties`, though this will be decided on a case by case basis depending on how much the web and api models differ.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
cburns93: I guess the case for the static final is a good one which definitely makes sense, however have you thought a way around this when testing? As you will not be able to mock the transformer response via Mockito.

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
cburns93: I'm just trying to understand this more but regarding your comment about 
>we'll be taking the balance sheet and converting it to a current period and to a previous period.

If you are to send data to the API, surely you would be sending the current period with its balance sheet data and not the balance sheet itself. Additionally, should we be converting balance sheet to period? I would assume its balanceSheet to balanceSheet?

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
BSaunders95: Good point. That's a problem which we've yet to solve, though for the purpose of this work, the static member feels like a logical approach and we'll work around mocking when we get to it.

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
cburns93: I'll retrieve my comment for now as I feel there is an understanding I'm probably missing in the web to api mapping.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
cburns93: :+1:

ljgeorge: Folks - why not make the getBalanceSheet method static and the call it directly BalanceSheetTransformer.getBalanceSheet() instead of having BalanceSheetTransformer as an instance member? Would this cause any thread safety issues?

cburns93: I don't think it poses any thread safety issues as static guarantees it will be thread safe. The bigger concern from my end is more the mocking of that method as it will not be possible via Mockito.

BSaunders95: Agreed Chris, though we shouldn't let unit tests define our code.

ljgeorge: I believe you can mock static classes using PowerMockito.  I think the uk.gov.ch.cap.server.sail.SailRegisterCommandTest in the chips repo is an example that mocks the use of the SailMarshaller class. Might be worth a look.

cburns93: >though we shouldn't let unit tests define our code.
In a general sense possibly not, however, the reason why TDD is popular is because it allows you to design the code in an efficient way which also allows you to test it. In this case I feel it may show a problem with our code which is what unit tests do, especially when done in a TDD way. However, there should be nothing wrong with writing the tests, you just won't be able to mock the transformer response, instead you will have to pass in a valid object for it to transform, which could raise the point of should we be also testing the transformer in the service class. The consensus normally is no, but like I said earlier, as long as we think about these things now instead of layer, we are sure to iron out as many potential problems as possible :)

Additionally, regarding the PowerMockito usage @ljgeorge, look above for reasons for why we are trying not to go into the PowerMock direction.

drosser: We avoid `PowerMock` where we can, but will we need to mock it? We initialise the Transformer as a static member, and it's sole purpose is to take an object of one type and turn it to another type. There's nothing intricate going on there, and nothing complicated. If we look at the code above, the `getBalanceSheet` method would, at full working order, take in 2 objects (Lets call them `Foo` and `Bar`), which it would get from the `API SDK`. The interaction with the SDK would be mocked, using something like `thenReturn` in `Mockito` we would be able to return a valid instantiated object of type `Foo` or `Bar` and use those in the static class.

ljgeorge: 👍 That's fine.

PR #8, opened 2018-08-02
Add baseLayout and fragments to support it

drosser:
- Add baseLayout
- Add betaBanner
- Add footer
- Add header
- Add userBar
- Add application properties to support links for view

Resolves: SFA-515.

### DIFF
--- a/src/main/resources/templates/fragments/betaBanner.html
+++ b/src/main/resources/templates/fragments/betaBanner.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+    <head>
+            <title></title>
+        </head>
+    <body>
+        <div th:fragment="betaBanner" class="phase-banner">
+            <p>
+                <strong class="phase-tag">BETA</strong>
+                <span>This is a trial service. Help us improve it by <a href="javascript:void(0)"><span class="visuallyhidden">This is a new service. Help us improve it by </span>completing our quick survey</a>.</span>
+            </p>
+        </div>
+    </body>
+</html>
--- a/src/main/resources/templates/fragments/betaBanner.html
+++ b/src/main/resources/templates/fragments/betaBanner.html
@@ -1,8 +1,8 @@
 <!DOCTYPE html>
 <html xmlns:th="http://www.thymeleaf.org">
     <head>
-            <title></title>
-        </head>
+        <title></title>
+    </head>
     <body>
         <div th:fragment="betaBanner" class="phase-banner">
             <p>
--- a/src/main/resources/templates/fragments/header.html
+++ b/src/main/resources/templates/fragments/header.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+    <head>
+        <title>Header</title>
+    </head>
+    <body>
+        <th:block th:fragment="header">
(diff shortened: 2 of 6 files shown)


src/main/resources/templates/fragments/betaBanner.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+    <head>
+            <title></title>
```
cburns93: empty title?

src/main/resources/templates/fragments/betaBanner.html
```diff
@@ -0,0 +1,14 @@
...
+        </head>
```
cburns93: misaligned ending header tag?

src/main/resources/templates/fragments/header.html
```diff
@@ -0,0 +1,31 @@
...
+                        <div class="content">
+                            <a href="#proposition-links" class="js-header-toggle menu">Menu</a>
+                            <nav id="proposition-menu">
+                                <a href="/" id="proposition-name"> File company accounts </a>
+                            </nav>
+                        </div>
+                    </div>
+                </div>
+            </header>
+            <div id="global-header-bar"></div>
```
cburns93: empty div?

src/main/resources/templates/fragments/userBar.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+    <head>
+        <title>User Bar</title>
+    </head>
+    <body>
+        <div th:fragment="userBar" class="js-toggle-nav" id="global-nav">
+            <a href="#navigation" id="signed-in-user-href" class="js-header-toggle" text="jsmith@testcompany.co.uk:"></a>
```
cburns93: hardcoded text?

src/main/resources/templates/fragments/userBar.html
```diff
@@ -0,0 +1,20 @@
...
+            <nav class="content" role="navigation">
+                <ul id="navigation" class="mobile-hidden">
+                    <li class="user" id="signed-in-user">jsmith@testcompany.co.uk: </li>
```
cburns93: hardcoded signed in user email?

cburns93:
Although you may not have addressed this in this PR, are you thinking about PIWIK yet? It may be worth thinking about it sooner rather than later!

src/main/resources/templates/fragments/betaBanner.html
BSaunders95: It's standard practise with Thymeleaf fragments to keep valid html even if it's empty so that the fragment can be rendered on its own.

src/main/resources/templates/fragments/userBar.html
BSaunders95: Just a placeholder, will be replaced with model attributes in due course :)

src/main/resources/templates/fragments/userBar.html
BSaunders95: As above

BSaunders95:
Piwik has been considered, but the way in which we interact with it will need to be reworked so it's been left out for now.

cburns93:
Approved, but forgot to comment about making sure we comment why we have empty tags, this is so no wally comes along in future and removes them because they are empty :+1:

PR #15, opened 2018-08-07
Add Balance Sheet Controller

BSaunders95:
- Add balance sheet controller. 
- Add base unit tests for the balance sheet controller.
- Update balance sheet service interface to take 2 parameters: transactionId and companyAccountsId.

Required for SFA-43

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
@@ -1,4 +1,49 @@
 package uk.gov.companieshouse.web.accounts.controller.smallfull;
 
+import javax.validation.Valid;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.ModelAttribute;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
+import uk.gov.companieshouse.web.accounts.service.smallfull.BalanceSheetService;
+
+@Controller
 public class BalanceSheetController {
+
+    private static final String BALANCE_SHEET_PATH = "/company/{companyNumber}/transaction/{transactionId}/company-accounts/{companyAccountsId}/small-full/balance-sheet";
+
+    private static final String SMALL_FULL_BALANCE_SHEET = "smallfull/balanceSheet";
+
+    @Autowired
+    private BalanceSheetService balanceSheetService;
+
+    @RequestMapping(value = BALANCE_SHEET_PATH, method = RequestMethod.GET)
+    public String getBalanceSheet(@PathVariable String transactionId,
+                                  @PathVariable String companyAccountsId,
+                                  Model model) {
+
+        model.addAttribute(balanceSheetService.getBalanceSheet(transactionId, companyAccountsId));
+
+        return SMALL_FULL_BALANCE_SHEET;
+    }
+
+    @RequestMapping(value = BALANCE_SHEET_PATH, method = RequestMethod.POST)
+    public String postBalanceSheet(@PathVariable String companyNumber,
+                                   @PathVariable String transactionId,
+                                   @PathVariable String companyAccountsId,
+                                   @ModelAttribute("balanceSheet") @Valid BalanceSheet balanceSheet,
+                                   BindingResult bindingResult) {
+
+        if (bindingResult.hasErrors()) {
+            return SMALL_FULL_BALANCE_SHEET;
+        }
+
+        // The next page will be returned once created
+        return SMALL_FULL_BALANCE_SHEET;
+    }
 }
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
@@ -5,10 +5,10 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.ModelAttribute;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
```diff
@@ -1,4 +1,49 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
+import uk.gov.companieshouse.web.accounts.service.smallfull.BalanceSheetService;
+
+@Controller
 public class BalanceSheetController {
-}
+
+    private static final String BALANCE_SHEET_PATH = "/company/{companyNumber}/transaction/{transactionId}/company-accounts/{companyAccountsId}/small-full/balance-sheet";
```
cburns93: So, back in Abridged, these URLs where flagged by Sonar because they were pasted in each controller. Could we have a function that creates these URLs dynamically? I'm only saving you the hassle of when Sonar raises it :(

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
```diff
@@ -1,4 +1,49 @@
...
+
+    private static final String SMALL_FULL_BALANCE_SHEET = "smallfull/balanceSheet";
+
+    @Autowired
+    private BalanceSheetService balanceSheetService;
+
+    @RequestMapping(value = BALANCE_SHEET_PATH, method = RequestMethod.GET)
```
cburns93: This can be changed from `@RequestMapping` to use the Get one instead `@GetMapping` which means you won't need the `RequestMethod.GET` anymore :)

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
```diff
@@ -1,4 +1,49 @@
...
+    public String getBalanceSheet(@PathVariable String transactionId,
+                                  @PathVariable String companyAccountsId,
+                                  Model model) {
+
+        model.addAttribute(balanceSheetService.getBalanceSheet(transactionId, companyAccountsId));
```
cburns93: Do you want to just pass the object through itself? Or do you want to give it a name? I assume it automatically takes the name from the object returned from `.getBalanceSheet` .. but is it worth giving a name for clarity and also for those who can just do a quick CMD + F for the name we've given it?

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
```diff
@@ -1,4 +1,49 @@
...
+
+        return SMALL_FULL_BALANCE_SHEET;
+    }
+
+    @RequestMapping(value = BALANCE_SHEET_PATH, method = RequestMethod.POST)
```
cburns93: This can be changed from `@RequestMapping` to use the Post one instead `@PostMapping` which means you won't need the `RequestMethod.POST` anymore :)

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String COMPANY_ACCOUNTS_ID = "companyAccountsId";
+
+    private static final String BALANCE_SHEET_PATH = "/company/" + COMPANY_NUMBER +
+                                                     "/transaction/" + TRANSACTION_ID +
+                                                     "/company-accounts/" + COMPANY_ACCOUNTS_ID +
+                                                     "/small-full/balance-sheet";
+
+    @BeforeEach
+    private void setup() {
+        MockitoAnnotations.initMocks(this);
```
cburns93: You shouldn't need this. If we are using JUnit5, which I believe we are, we just need to annotate the class `@ExtendWith(MockitoExtension.class)`

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
```diff
@@ -0,0 +1,69 @@
...
+        this.mockMvc = MockMvcBuilders.standaloneSetup(controller).build();
+    }
+
+    @Test
+    void getRequestSuccess() throws Exception {
```
cburns93: Probably worth adding `@DisplayName` to the test to just give a nice description what it does.

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
```diff
@@ -0,0 +1,69 @@
...
+        when(balanceSheetService.getBalanceSheet(TRANSACTION_ID, COMPANY_ACCOUNTS_ID)).thenReturn(new BalanceSheet());
+
+        this.mockMvc.perform(get(BALANCE_SHEET_PATH))
+                    .andExpect(status().isOk())
+                    .andExpect(view().name("smallfull/balanceSheet"))
+                    .andExpect(model().attributeExists("balanceSheet"));
+    }
+
+    @Test
+    void postRequestSuccess() throws Exception {
```
cburns93: Probably worth adding `@DisplayName` to the test to just give a nice description what it does.

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    void getRequestSuccess() throws Exception {
+
+        when(balanceSheetService.getBalanceSheet(TRANSACTION_ID, COMPANY_ACCOUNTS_ID)).thenReturn(new BalanceSheet());
+
+        this.mockMvc.perform(get(BALANCE_SHEET_PATH))
+                    .andExpect(status().isOk())
+                    .andExpect(view().name("smallfull/balanceSheet"))
```
cburns93: We could assign the view name and attribute name to a global test variable for other tests to use? instead of the string literals

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
BSaunders95: We're looking into implementing an interface in each controller to construct previous / next paths to redirect between each page of the journey so we'll look to resolve it there. Cheers for raising it 👍

cburns93: 👍

PR #16, opened 2018-08-08
Add Session Handler to the Filter Chain

BSaunders95:
Add session handler to the request filter chain.
Add user details interceptor which executes before a view renders to retrieve the email from the session and append it to the model.
Update the user bar fragment to display the email from the model.

Required for SFA-43

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
@@ -1,12 +1,28 @@
 package uk.gov.companieshouse.web.accounts;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+import uk.gov.companieshouse.web.accounts.interceptor.UserDetailsInterceptor;
 
 @SpringBootApplication
-public class CompanyAccountsWebApplication {
+public class CompanyAccountsWebApplication implements WebMvcConfigurer {
+
+    UserDetailsInterceptor userDetailsInterceptor;
+
+    @Autowired
+    public CompanyAccountsWebApplication(UserDetailsInterceptor userDetailsInterceptor) {
+        this.userDetailsInterceptor = userDetailsInterceptor;
+    }
 
     public static void main(String[] args) {
         SpringApplication.run(CompanyAccountsWebApplication.class, args);
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
```diff
@@ -1,12 +1,28 @@
...
+        this.userDetailsInterceptor = userDetailsInterceptor;
+    }
 
     public static void main(String[] args) {
         SpringApplication.run(CompanyAccountsWebApplication.class, args);
     }
+
+    @Override
+    public void addInterceptors(InterceptorRegistry registry) {
+        registry.addInterceptor(userDetailsInterceptor);
```
macreman: does this automatically add this interceptor to every route? I've seen this done with `.addPathPatterns("/**")`

https://github.com/companieshouse/transactions.api.ch.gov.uk/blob/develop/src/main/java/uk/gov/companieshouse/api/transactions/TransactionsApplication.java#L82

BSaunders95: It does work for all routes, but `addPathPatterns` can be added to make it more explicit. Typically `addPathPatterns` should be used to exclude certain routes.

src/main/java/uk/gov/companieshouse/web/accounts/interceptor/UserDetailsInterceptor.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public void postHandle(HttpServletRequest request, HttpServletResponse response, Object handler, @Nullable ModelAndView modelAndView) throws Exception {
+
+        if (modelAndView != null) {
+            Map<String, Object> sessionData = SessionHandler.getSessionDataFromContext();
+            Map<String, Object> signInInfo = (Map<String, Object>) sessionData.get(SIGN_IN_KEY);
+            if (signInInfo != null) {
+                Map<String, Object> userProfile = (Map<String, Object>) signInInfo.get(USER_PROFILE_KEY);
+                modelAndView.getModelMap().addAttribute(USER_EMAIL, userProfile.get(EMAIL_KEY));
```
jsimons1989: What is the user's email used for? What is the expected behaviour if the email is missing?

BSaunders95: The email is displayed in the user nav bar. The session must exist for the page to render, otherwise we force a redirect to re-authenticate. So theoretically, there won't ever be an issue in retrieving the email (unless there's a drastic change where the structure of the session changes). Also, logging any of the session data should be done cautiously.

BSaunders95: There could be a conditional statement added to the Thymeleaf fragment to hide the span if the email is null?

PR #17, opened 2018-08-08
Prefer constructor injection for mandatory dependencies

marcransome:
As per the Spring documentation on the subject, constructor injection should be preferred for mandatory dependencies.

Reference: https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#beans-constructor-injection

Resolves: SFA-43


jsimons1989:
For info, with autowiring there are more ways to initialise variables:

Field injection
```java
@Autowired
private BalanceSheetController controller
```

Constructor injection
```java
@Autowired
public BalanceSheetController(BalanceSheetService balanceSheetService) {
   this.balanceSheetService = balanceSheetService;
}
```

Setter injection
```java
@Autowired
public void setBalanceSheetController(BalanceSheetService balanceSheetService) {
   this.balanceSheetService = balanceSheetService;
}
```

PR #24, opened 2018-08-13
Add Transaction Api Interaction to Create a Transaction

BSaunders95:
Required for SFA-123

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
@@ -1,17 +1,31 @@
 package uk.gov.companieshouse.web.accounts.service.transaction.impl;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
```diff
@@ -1,18 +1,32 @@
...
+import uk.gov.companieshouse.api.error.ApiErrorResponseException;
 import uk.gov.companieshouse.api.model.transaction.Transaction;
+import uk.gov.companieshouse.web.accounts.api.ApiClientService;
 import uk.gov.companieshouse.web.accounts.service.transaction.TransactionService;
 
 @Service
 public class TransactionServiceImpl implements TransactionService {
 
+    @Autowired
+    ApiClientService apiClientService;
```
ChrisJBurns: Im guessing this class is a temporary thing until we do the private sdk?

BSaunders95: No, this service layer will handle transaction api requests via the SDK :)

PR #26, opened 2018-08-14
Add SDK to Get Current Period from the API

BSaunders95:
Adds use of the SDK to retrieve the current period for the balance sheet.

Note that once approved, this branch will only be merged following the merge of PR: https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/41 which exposes the GET endpoint in the API.

Required for SFA-596

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
@@ -10,6 +10,7 @@ import org.springframework.web.bind.annotation.ModelAttribute;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
+import uk.gov.companieshouse.api.error.ApiErrorResponseException;
 import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
 import uk.gov.companieshouse.web.accounts.service.smallfull.BalanceSheetService;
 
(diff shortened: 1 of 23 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -12,14 +16,22 @@
     @Autowired
     private BalanceSheetTransformer transformer;
 
+    @Autowired
+    private ApiClientService apiClientService;
+
     @Override
-    public BalanceSheet getBalanceSheet(String transactionId, String companyAccountsId) {
+    public BalanceSheet getBalanceSheet(String transactionId, String companyAccountsId)
+            throws ApiErrorResponseException {
```
jsimons1989: This is inconsistent with how we are handling errors in the [API services](https://github.com/companieshouse/company-accounts.api.ch.gov.uk/blob/develop/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java). We have separated the knowledge bleed between layer so the controller doesn't know the internal behaviour of the service (making an API call).

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
```diff
@@ -27,7 +28,12 @@ public String getBalanceSheet(@PathVariable String transactionId,
                                   @PathVariable String companyAccountsId,
                                   Model model) {
 
-        model.addAttribute("balanceSheet", balanceSheetService.getBalanceSheet(transactionId, companyAccountsId));
+        try {
+            model.addAttribute("balanceSheet", balanceSheetService.getBalanceSheet(transactionId, companyAccountsId));
+        } catch (ApiErrorResponseException e) {
+            // TODO: Handle with exception handler once implemented
+            e.printStackTrace();
```
jsimons1989: This should be using structured logging (if it gets missed in a future PR it won't be logging in the correct format).

BSaunders95: Used the Structured Logging library for `LOGGER.error()` in place of printing the stack trace. Until Api validation is enforced, we can assume these errors are truly exceptional so the `error` log level would be the correct choice as opposed to `info` or similar.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
BSaunders95: As discussed, we'll address this in a meeting with everyone. I'm of the opinion that we should be following the 'throw early, catch late' principle, but I'm happy to put it to the team

PR #27, opened 2018-08-14
Add Current Period to Balance Sheet Transformer

BSaunders95:
Add current period as a parameter in the `getBalanceSheet` method of the balance sheet transformer.

Note: this PR is set up against https://github.com/companieshouse/company-accounts.web.ch.gov.uk/pull/26

Required for SFA-592

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
@@ -1,6 +1,7 @@
 package uk.gov.companieshouse.web.accounts.transformer.smallfull.impl;
 
 import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.model.accounts.smallfull.CurrentPeriod;
 import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
 import uk.gov.companieshouse.web.accounts.model.smallfull.CalledUpShareCapitalNotPaid;
 import uk.gov.companieshouse.web.accounts.transformer.smallfull.BalanceSheetTransformer;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
```diff
@@ -1,17 +1,24 @@
...
 import uk.gov.companieshouse.web.accounts.model.smallfull.CalledUpShareCapitalNotPaid;
 import uk.gov.companieshouse.web.accounts.transformer.smallfull.BalanceSheetTransformer;
 
 @Component
 public class BalanceSheetTransformerImpl implements BalanceSheetTransformer {
 
-    public BalanceSheet getBalanceSheet() {
+    public BalanceSheet getBalanceSheet(CurrentPeriod currentPeriod) {
+
+        uk.gov.companieshouse.api.model.accounts.smallfull.BalanceSheet apiCurrentPeriodBalanceSheet = currentPeriod.getBalanceSheet();
```
ChrisJBurns: Can we avoid the ambiguous paths?

BSaunders95: The only way around this really is to rename the models in the SDK, which I'm more than happy to do if we can agree a standard :)

ChrisJBurns: I think we agreed that the API shouldn't have a prefix/suffix name, but the web should as we're the only ones viewing it. So it becomes `BalanceSheetWeb` as the API model is the "source of truth regarding models". Additionally, we are very early in the days of the web so it wouldn't be any effort to change :)

ChrisJBurns: Furthermore, I did start doing this change in abridged but had to stop because it would have meant renaming things in the regression tests. So it would be good to do it from the start.

BSaunders95: Discussed with the team; we opted to suffix SDK models with 'Api'

PR #36, opened 2018-08-28
Change Return Type of createCompanyAccounts

BSaunders95:
Alter parameter types and return type of the createCompanyAccounts method for encapsulation of knowledge.

Required for SFA-573

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/companyaccounts/CompanyAccountsService.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/companyaccounts/CompanyAccountsService.java
@@ -1,10 +1,11 @@
 package uk.gov.companieshouse.web.accounts.service.companyaccounts;
 
+import com.google.api.client.util.DateTime;
 import java.util.Date;
 import uk.gov.companieshouse.api.error.ApiErrorResponseException;
 import uk.gov.companieshouse.api.model.accounts.CompanyAccountsApi;
 
 public interface CompanyAccountsService {
 
-    CompanyAccountsApi createCompanyAccounts(String transactionId, Date periodEndOn) throws ApiErrorResponseException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/companyaccounts/CompanyAccountsService.java
```diff
@@ -1,10 +1,9 @@
...
 
-import java.util.Date;
+import com.google.api.client.util.DateTime;
 import uk.gov.companieshouse.api.error.ApiErrorResponseException;
-import uk.gov.companieshouse.api.model.accounts.CompanyAccountsApi;
 
 public interface CompanyAccountsService {
 
-    CompanyAccountsApi createCompanyAccounts(String transactionId, Date periodEndOn) throws ApiErrorResponseException;
+    String createCompanyAccounts(String transactionId, DateTime periodEndOn) throws ApiErrorResponseException;
```
ChrisJBurns: Why are we changing the datetime type? We have had problems in the past where there where multiple types of DateTime, it may cause problems in the future about this as the model has the Java8 Date but now we are changing this to accept google date?

BSaunders95: It's Google's `DateTime` because that's the type returned in the model from the SDK (The JSON parser used in the SDK only operates on primitives... and Google's custom types). It seems pointless to take the `DateTime` out of the Company Profile model, convert it to a `LocalDate`, pass it through the company accounts service, and convert it back to a `DateTime` for use in the Company Accounts model.

BSaunders95: So when the models had a `Date` field, the JSON parser threw an exception because `Date` is not a primitive type. Interestingly, `String`s are considered primitives by the parser. Unfortunately our hand is forced on this occasion; we have to use `DateTime`.

PR #42, opened 2018-08-28
Add Service Method to Post a Balance Sheet

BSaunders95:
Add service layer functionality to post the balance sheet.
Add to balance sheet transformer to turn a web balance sheet model into an api current period model.

SFA-572

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
@@ -35,4 +35,18 @@ public class BalanceSheetServiceImpl implements BalanceSheetService {
 
         return balanceSheet;
     }
+
+    @Override
+    public void postBalanceSheet(String transactionId, String companyAccountsId, BalanceSheet balanceSheet)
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -35,4 +35,18 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
+            throws ApiErrorResponseException {
+
+        ApiClient apiClient = apiClientService.getApiClient();
+
+        CurrentPeriodApi currentPeriod = transformer.getCurrentPeriod(balanceSheet);
+
+        apiClient.transaction(transactionId)
+                .companyAccount(companyAccountsId)
+                .smallFull()
+                .currentPeriod().create(currentPeriod);
```
paulforsyth: do you not care about any return values from the create call?

BSaunders95: Not particularly; if any errors occur an exception will be thrown and caught higher up anyway :)

PR #45, opened 2018-08-31
Add structured logging interceptor

marcransome:
Add structured logging interceptor, dependencies and unit tests.

Resolves: SFA-597

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
         <thymeleaf-layout-dialect.version>2.3.0</thymeleaf-layout-dialect.version>
 
         <hamcrest.version>1.3</hamcrest.version>
+        <hamcrest-library-version>1.3</hamcrest-library-version>
 
         <java-session-handler.version>2.0.0-rc4</java-session-handler.version>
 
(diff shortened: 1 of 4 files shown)


Mathsh:
Guru mentioned the need for an @bean LoggerApi when using the structured logging, which can be seen in the api usage. Is this still the case?

marcransome:
This was removed as part of a refactoring to use the logging helpers and is no longer required: https://github.com/companieshouse/company-accounts.api.ch.gov.uk/commit/a9ba4e122190eb0bdea2b220daf16cd176a99fda#diff-06bcd4347458c6d3e9945bf1c95074bbL82

pom.xml
```diff
@@ -134,13 +135,27 @@
             <artifactId>spring-security-config</artifactId>
         </dependency>
 
+        <!-- Logging dependencies -->
+        <dependency>
+            <groupId>uk.gov.companieshouse</groupId>
+            <artifactId>structured-logging</artifactId>
+            <version>1.4.0-rc2</version>
```
ljgeorge: Should we put the  structured logging version into a property?

PR #46, opened 2018-09-03
Log Controller Errors with Request Id

BSaunders95:
Use `errorRequest` to log errors in each controller so logs include the requestId

Resolves SFA-597

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.web.accounts.controller.smallfull;
 
+import javax.servlet.http.HttpServletRequest;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
```diff
@@ -28,13 +29,14 @@ public String getApproval() {
...
-                               @PathVariable String companyAccountsId) {
+                               @PathVariable String companyAccountsId,
+                               HttpServletRequest request) {
 
         try {
             transactionService.closeTransaction(transactionId);
         } catch (ApiErrorResponseException e) {
             // TODO: handle ApiErrorResponseExceptions (SFA-594)
-            LOGGER.error(e);
+            LOGGER.errorRequest(request, "Failed to close transaction");
```
macreman: We should be passing the exception to the logger as it may contain extra details that help us identify why the service "Failed to close transaction"
This might require an extra method being exposed in the structured-logging library to allow an exception and message at the same time but thats better than losing the specific details of the exception (stack trace and cause exception)

PR #47, opened 2018-09-04
Add exception handling and update unit tests

marcransome:
Add suitable exception handling—wrapping lower level service-layer exceptions in a new `ServiceException` checked exception—and update unit tests with failure paths for error page returned to user agent.

Resolves: SFA-594

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
@@ -7,8 +7,8 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
-import uk.gov.companieshouse.api.error.ApiErrorResponseException;
 import uk.gov.companieshouse.web.accounts.controller.BaseController;
+import uk.gov.companieshouse.web.accounts.exception.ServiceException;
 import uk.gov.companieshouse.web.accounts.service.transaction.TransactionService;
 
 @Controller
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/TransactionService.java
```diff
@@ -1,11 +1,10 @@
 package uk.gov.companieshouse.web.accounts.service.transaction;
 
-import uk.gov.companieshouse.api.error.ApiErrorResponseException;
-import uk.gov.companieshouse.api.model.transaction.Transaction;
+import uk.gov.companieshouse.web.accounts.exception.ServiceException;
 
 public interface TransactionService {
 
-    String createTransaction(String companyNumber) throws ApiErrorResponseException;
+    String createTransaction(String companyNumber) throws ServiceException;
```
ChrisJBurns: could you have java doc on these methods? and then inherit them in the implementation?

src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/TransactionService.java
```diff
@@ -1,11 +1,10 @@
...
 
-    void closeTransaction(String transactionId) throws ApiErrorResponseException;
+    void closeTransaction(String transactionId) throws ServiceException;
```
ChrisJBurns: same here

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
```diff
@@ -34,9 +34,10 @@ public String postApproval(@PathVariable String companyNumber,
 
         try {
             transactionService.closeTransaction(transactionId);
-        } catch (ApiErrorResponseException e) {
-            // TODO: handle ApiErrorResponseExceptions (SFA-594)
+        } catch (ServiceException e) {
+
             LOGGER.errorRequest(request, "Failed to close transaction", e);
+            return "error";
```
BSaunders95: Could we have a protected string in the base controller to hold the error view?

src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/TransactionService.java
BSaunders95: Do we need Javadocs here? Only from the perspective that our service methods aren't exposed and the functionality should be easily deduced from the method names.

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
marcransome: I've added a suitable string constant to the base controller and replaced the string literals in the controllers with this.

PR #50, opened 2018-09-05
Add unit test for user details interceptor and interface to support test

marcransome:
This change introduces a unit test for the `UserDetailsInterceptor` class, to improve code coverage, as well as a convenience interface that simplifies testing. The interface is necessary since a static method—`getSessionDataFromContext()`—is being used, and to avoid imposing the use of a test framework for mocking static methods.


BSaunders95:
I believe it's a JUnit 5 thing but we have complaints making tests `public`.

ChrisJBurns:
@rafahop  Indeed what Brad said. Access modifiers on test methods are now not needed with JUnit 5. 

> _"The most obvious change is that test classes and methods do not have to be public anymore. Package visibility suffices but private does not. I think this is a sensible choice and in line with how we intuit the different visibility modifiers."_

rafahop:
Yes, I know. That is what I was highlighting. That new tests in this PR are public whereas this repo seem to keep them default. I particularly don't see the benefit though

BSaunders95:
Ah I see, sorry. I've removed the public modifiers for consistency 👍

PR #51, opened 2018-09-05
Resolve sonar code smells

marcransome:
These changes resolve several code smells highlighted by Sonar:

* Make base controller constructor private
* Remove unnecessary `TODO` comment


src/main/java/uk/gov/companieshouse/web/accounts/controller/BaseController.java
```diff
@@ -6,6 +6,8 @@
 
 public abstract class BaseController {
 
+    protected BaseController() { }
```
rafahop: Can the constructor be placed below the constants definition please?

PR #58, opened 2018-09-06
Add Back Page to Models

BSaunders95:
Add back page link as a model attribute for display on all screens which have a back link


BSaunders95:
The back link is an `href` link to the previous screen in the linear journey. The method to add the attribute is only called currently in `BalanceSheetController` and `ApprovalController` since `StepsToCompleteController` doesn't have a screen 'behind' it.

PR #61, opened 2018-09-10
Fix UserDetailsInterceptor

BSaunders95:
Execute user details interceptor for POST requests which don't redirect so that user email is displayed on the error page.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/interceptor/UserDetailsInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/interceptor/UserDetailsInterceptor.java
@@ -26,9 +26,8 @@ public class UserDetailsInterceptor extends HandlerInterceptorAdapter {
     @Override
     public void postHandle(HttpServletRequest request, HttpServletResponse response, Object handler, @Nullable ModelAndView modelAndView) throws Exception {
 
-        if ((request.getMethod().equalsIgnoreCase("GET") ||
-             request.getMethod().equalsIgnoreCase("POST") && !modelAndView.getViewName().startsWith("redirect:"))
-                && modelAndView != null) {
+        if (modelAndView != null && (request.getMethod().equalsIgnoreCase("GET") ||
+             request.getMethod().equalsIgnoreCase("POST") && !modelAndView.getViewName().startsWith("redirect:"))) {
 
             Map<String, Object> sessionData = sessionService.getSessionDataFromContext();
             Map<String, Object> signInInfo = (Map<String, Object>) sessionData.get(SIGN_IN_KEY);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/interceptor/UserDetailsInterceptor.java
```diff
@@ -26,7 +27,9 @@
     @Override
     public void postHandle(HttpServletRequest request, HttpServletResponse response, Object handler, @Nullable ModelAndView modelAndView) throws Exception {
 
-        if (request.getMethod().equalsIgnoreCase("GET") && modelAndView != null) {
+        if (modelAndView != null && (request.getMethod().equalsIgnoreCase("GET") ||
```
ChrisJBurns: So my understanding of this was slightly confusing when I initially looked at it and I believe you may be able to make this more readable by formatting it like the following.
```java
if (modelAndView != null && !modelAndView.getViewName().startsWith(UrlBasedViewResolver.REDIRECT_URL_PREFIX)
&& (request.getMethod().equalsIgnoreCase("GET") ||
request.getMethod().equalsIgnoreCase("POST"))

BSaunders95: In my opinion, the conditional check that method equals `POST` and view name doesn't start with `redirect:` should be compounded incase at some point a redirect is introduced in a `GET`

ChrisJBurns: I'm not really sure I understand the point you're trying to make here, my comment was only regarding the order in which you do them. My confusion came from where you are checking for null on the `modelAndView` and then you're doing another check against the `getMethod` and then checking the `modelAndView` again. It would be much easier to group both of the `modelAndView` checks into one line, thus making it easier for the developer to know what checks you're doing. Otherwise you unintentionally give off the impression that the second `modelAndView` check is part of another condition.

ChrisJBurns: Ahh okay, that makes it easier. I think it was including the second modelView checking all the time which is why I recommended the first solution. The only other comment I have is that you could possible at a predicate here to make it easier to read something like 
```java
Predicate<String> getMethod = e -> request.getMethod().equalsIgnoreCase(e);``` and used like
```java
(getMethod.equals("GET") ||
                (getMethod.equals("POST") ...
```but I'm not really arsy on that tbh, I'm only spitballing. (that may need editing slightly - wrote of top of head)

BSaunders95: I'll leave it for now since we only return the method name twice, I don't think it's too much duplication. Nice suggestion though!

PR #62, opened 2018-09-11
Small Full Criteria Page

BSaunders95:
- Add `CriteriaController` and corresponding model and view for the small full criteria page
- Add back button link to the Steps to Complete screen now that a page precedes it

Required for SFA-111

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/CriteriaController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/CriteriaController.java
@@ -10,13 +10,14 @@ import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import uk.gov.companieshouse.web.accounts.annotation.NextController;
+import uk.gov.companieshouse.web.accounts.controller.BaseController;
 import uk.gov.companieshouse.web.accounts.model.smallfull.Criteria;
 import uk.gov.companieshouse.web.accounts.util.Navigator;
 
 @Controller
 @NextController(StepsToCompleteController.class)
 @RequestMapping("/company/{companyNumber}/small-full/criteria")
-public class CriteriaController {
+public class CriteriaController extends BaseController {
 
     private static final String TEMPLATE = "smallfull/criteria";
 
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/CriteriaController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/CriteriaController.java
@@ -1,0 +1,48 @@
+package uk.gov.companieshouse.web.accounts.controller.smallfull;
+
+import javax.validation.Valid;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.ModelAttribute;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import uk.gov.companieshouse.web.accounts.annotation.NextController;
+import uk.gov.companieshouse.web.accounts.model.smallfull.Criteria;
+import uk.gov.companieshouse.web.accounts.util.Navigator;
+
+@Controller
+@NextController(StepsToCompleteController.class)
+@RequestMapping("/company/{companyNumber}/small-full/criteria")
+public class CriteriaController {
+
+    private static final String TEMPLATE = "smallfull/criteria";
+
+    @GetMapping
+    public String getCriteria(Model model) {
+
+        Criteria criteria = new Criteria();
+        model.addAttribute("criteria", criteria);
+
+        return TEMPLATE;
+    }
+
+    @PostMapping
+    public String postCriteria(@PathVariable String companyNumber,
+                               @ModelAttribute("criteria") @Valid Criteria criteria,
+                               BindingResult bindingResult) {
+
+        if (bindingResult.hasErrors()) {
+            return TEMPLATE;
+        }
+
+        if (!criteria.getIsCriteriaMet()) {
+            // Temporarily return the criteria page until a 'criteria not met' page is created
+            return TEMPLATE;
+        }
+
+        return Navigator.getNextControllerRedirect(this.getClass(), companyNumber);
+    }
+}
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -1,6 +1,9 @@
 # Provides a 'catch all' for errors when casting text inputs from strings to integers
 typeMismatch.java.lang.Integer=Enter a value without commas, brackets or other symbols
 
+# Accounts criteria
+isCriteriaMet=
+
 # Balance Sheet
 # Called Up Share Capital
 calledUpShareCapitalNotPaid.currentAmount=Called up share capital not paid current period:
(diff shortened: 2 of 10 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/CriteriaController.java
```diff
@@ -0,0 +1,49 @@
...
+    public String postCriteria(@PathVariable String companyNumber,
+                               @ModelAttribute("criteria") @Valid Criteria criteria,
+                               BindingResult bindingResult) {
+
+        if (bindingResult.hasErrors()) {
+            return TEMPLATE;
+        }
+
+        if (!criteria.getIsCriteriaMet()) {
+            // Temporarily return the criteria page until a 'criteria not met' page is created
```
marcransome: It may be beneficial to mark this comment as `TODO` to ensure this is picked up by Sonar and not missed in future. changes

src/test/java/uk/gov/companieshouse/web/accounts/exception/GlobalExceptionHandlerTests.java
```diff
@@ -41,8 +40,6 @@ private void setup() {
     @DisplayName("Global Exception Handler Test - Assert runtime exceptions are caught")
     public void testGlobalExceptionHandlerError() throws Exception {
 
-        when(controller.getStepsToComplete()).thenThrow(new RuntimeException());
-
```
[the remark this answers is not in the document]
BSaunders95: The `getStepsToComplete` method now takes 2 parameters - a companyNumber and the model; using `any` matchers on the types, Mockito complains about unnecessary stubbings - apparently this call is not required and a RuntimeException is thrown regardless. Not sure how to proceed.

src/main/resources/messages.properties
```diff
@@ -1,6 +1,9 @@
 # Provides a 'catch all' for errors when casting text inputs from strings to integers
 typeMismatch.java.lang.Integer=Enter a value without commas, brackets or other symbols
 
+# Accounts criteria
+isCriteriaMet=
+
```
marcransome: Is this message used anywhere?

BSaunders95: Yes, it's used when a binding error is triggered on `POST` calls in the global errors section. We don't display the name of the field, hence it's left as an empty string. Without this, Thymeleaf attempts to resolve the field name without success and the display isn't correct.

jsimons1989: Can this be handled in the template instead of adding empty properties (making the assumption there could be more occurrences of this)? If not, can you add a comment to explain why it's empty please.

src/main/resources/templates/layouts/baseLayout.html
```diff
@@ -18,7 +18,7 @@
         th:href="@{{cdnUrl}/images/favicon.ico?0.17.3(cdnUrl=${@environment.getProperty('cdn.url')})}"
         type="image/x-icon" />
 </head>
-<body>
+<body class="">
```
marcransome: Is the class attribute here required?

jsimons1989:
I wasn't able to display this code in my IDE without compilation errors in the classes where Lombok was used. The getters and setters were not found.

src/main/resources/templates/layouts/baseLayout.html
```diff
@@ -29,5 +29,10 @@
         </main>
     </div>
     <div th:replace="fragments/footer :: footer"></div>
+    <script type="text/javascript">
```
jsimons1989: The empty body class requiring this JS block could get further separated from this block as more features are added to this layout HTML page. This could be a pain for a developer to figure out.

Also, our JS is in JS files in the CDN repository, we should have a consistent approach.

src/main/resources/templates/smallfull/criteria.html
```diff
@@ -0,0 +1,85 @@
...
+            <ul class="list list-bullet text" id="criteria-additional-criteria-list">
+                <li>turnover is not over £10.2 million</li>
+                <li>balance sheet total is not over £5.1 million</li>
+                <li>average number of employees is not over 50</li>
+            </ul>
+
+            <h2 class="heading-medium" id="citeria-submit-heading">Do your prepared full accounts meet the criteria?</h2>
+
+            <fieldset class="form-group" th:classappend="${#fields.hasErrors('isCriteriaMet')} ? 'error' : ''">
+                <div class="form-group">
```
ch-mcabrera: We might need here the <legend> element, I remembered James Francis to mention something about it as it is used by some screen readers.

src/main/resources/templates/layouts/baseLayout.html
BSaunders95: Sorry, lifted from abridged. Will move this to cdn.

src/main/resources/templates/smallfull/criteria.html
BSaunders95: Which element?

src/main/resources/templates/smallfull/criteria.html
```diff
@@ -0,0 +1,85 @@
...
+
+            <ul class="list list-bullet text" id="criteria-additional-criteria-list">
+                <li>turnover is not over £10.2 million</li>
+                <li>balance sheet total is not over £5.1 million</li>
+                <li>average number of employees is not over 50</li>
+            </ul>
+
+            <h2 class="heading-medium" id="citeria-submit-heading">Do your prepared full accounts meet the criteria?</h2>
+
+            <fieldset class="form-group" th:classappend="${#fields.hasErrors('isCriteriaMet')} ? 'error' : ''">
```
ch-mcabrera: shouldnt the `fieldset` element wrapped by `form` element?

src/main/resources/templates/layouts/baseLayout.html
marcransome: It's also worth considering using jQuery to append the class in this case which would be more readable: `$('body').addClass('js-enabled');`

src/main/resources/templates/layouts/baseLayout.html
```diff
@@ -29,5 +29,10 @@
...
+        document.body.className = ((document.body.className) ? document.body.className
+                + ' js-enabled'
+                : 'js-enabled');
+    </script>
```
marcransome: It's worth considering using jQuery to append the class to the 'body' element in this case, which would be more readable: `$('body').addClass('js-enabled');`

src/test/java/uk/gov/companieshouse/web/accounts/exception/GlobalExceptionHandlerTests.java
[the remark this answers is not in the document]
marcransome: Could we remove the `GlobalExceptionHandlerTests.java` file? We have some concerns around this and can discuss separately.

src/main/resources/templates/smallfull/criteria.html
ch-mcabrera: `<legend>`, I've just checked the prototype and I noticed that this element was there.

src/main/resources/messages.properties
BSaunders95: Using a ternary operator to check if the property is null, Thymeleaf still falters for some reason and the global error is formatted as:
```
??<message>_en_US??
```
I'll add a comment for now for clarity

src/main/resources/templates/smallfull/criteria.html
marcransome: Isn't this a child element of the `<form>` on line 12?

src/main/resources/templates/layouts/baseLayout.html
BSaunders95: Apparently not, has been removed

src/main/resources/templates/smallfull/criteria.html
BSaunders95: Added, thanks

src/main/resources/templates/layouts/baseLayout.html
BSaunders95: This has been done, so this PR is now dependent on https://github.com/companieshouse/cdn.ch.gov.uk/pull/261

src/main/resources/templates/layouts/baseLayout.html
BSaunders95: Moved to cdn, PR: https://github.com/companieshouse/cdn.ch.gov.uk/pull/261

PR #70, opened 2018-09-24
Use static css for accounts web app styling

leahtaylor:
Add static css stylesheets for the accounts web incorporating new govuk styles.

Remove references to old cdn stylesheets

_*Temporary measure until cdn is updated with new gov uk stylesheets*_

SFA-336


drosser:
Just gotta get rid of the CSS file from in here then we look good to go

leahtaylor:
application.css is removed in PR https://github.com/companieshouse/company-accounts.web.ch.gov.uk/pull/71

PR #75, opened 2018-09-26
Add fixed assets section to balance sheet template

leahtaylor:
Add fixed assets section to balance sheet

Add fixed Assets object to balance sheet model
Create fixed assets class
Create tangible assets class

Add fixed assets section to template

SFA-336


src/main/resources/templates/smallfull/balanceSheet.html
```diff
@@ -7,7 +7,7 @@
     <head>
         <title>Balance Sheet</title>
     </head>
-
+<body>
```
drosser: Should this be here? The [baseLayout](https://github.com/companieshouse/company-accounts.web.ch.gov.uk/blob/develop/src/main/resources/templates/layouts/baseLayout.html#L18) is responsible for holding the body, and every page we develop has the [content](https://github.com/companieshouse/company-accounts.web.ch.gov.uk/blob/develop/src/main/resources/templates/layouts/baseLayout.html#L26) layout fragment to be added to the body

src/main/resources/templates/smallfull/balanceSheet.html
```diff
@@ -61,7 +61,37 @@ <h2 class="govuk-heading-s">
                         </div>
                     </div>
 
-                    <br>
+                    <!--Fixed Assets-->
+
+                    <h2 class="govuk-heading-m">Fixed assets</h2>
+
+                    <div class="govuk-grid-row">
```
BSaunders95: Should there be an error row above this?

leahtaylor: Was going to add the validation in my next PR so will add the error row then

PR #77, opened 2018-09-27
Add API validation error support

marcransome:
These changes introduce a number of custom annotations—namely `ValidationModel` and `ValidationMapping`—to support dynamic binding of API validation errors to model object fields.

The process works as follows:

- On application startup the models package is scanned for model classes annotated with `@ValidationModel`
- For each candidate class, the declared fields are scanned for `@ValidationMapping` annotations; primitive or `String` fields _without_ the annotation are ignored, other fields become candidates themselves for the same class scanning mechanism
- For each `@ValidationMapping` annotation, a mapping is created for the field name (introspected) and the full JSON error path taken from the annotation
- At runtime, API validation errors are mapped to fields using the `ValidationContext` class, and a support method added to the `BaseController` class

Significant classes and methods—both `public` and `private`—include Javadoc comments with explanatory notes on their function.

Resolves: SFA-706

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
@@ -3,22 +3,15 @@ package uk.gov.companieshouse.web.accounts;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
-import uk.gov.companieshouse.logging.Logger;
-import uk.gov.companieshouse.logging.LoggerFactory;
 import uk.gov.companieshouse.web.accounts.interceptor.LoggingInterceptor;
 import uk.gov.companieshouse.web.accounts.interceptor.UserDetailsInterceptor;
-import uk.gov.companieshouse.web.accounts.util.ValidationContext;
 
 @SpringBootApplication
 public class CompanyAccountsWebApplication implements WebMvcConfigurer {
 
     public static final String APPLICATION_NAME_SPACE = "company-accounts.web.ch.gov.uk";
-    public static final String APPLICATION_MODEL_PACKAGE = "uk.gov.companieshouse.web.accounts.model";
-
-    private static final Logger LOGGER = LoggerFactory.getLogger(APPLICATION_NAME_SPACE);
 
     private UserDetailsInterceptor userDetailsInterceptor;
     private LoggingInterceptor loggingInterceptor;
@@ -32,15 +25,6 @@ public class CompanyAccountsWebApplication implements WebMvcConfigurer {
     }
 
     public static void main(String[] args) {
-
-        try {
-            ClassPathScanningCandidateComponentProvider provider = new ClassPathScanningCandidateComponentProvider(true);
-            ValidationContext.scanPackageForValidationMappings(provider, APPLICATION_MODEL_PACKAGE);
-        } catch (RuntimeException e) {
-            LOGGER.error(e.getMessage());
-            System.exit(1);
-        }
-
         SpringApplication.run(CompanyAccountsWebApplication.class, args);
     }
 
--- a/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
@@ -3,15 +3,22 @@ package uk.gov.companieshouse.web.accounts;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
 import uk.gov.companieshouse.web.accounts.interceptor.LoggingInterceptor;
 import uk.gov.companieshouse.web.accounts.interceptor.UserDetailsInterceptor;
+import uk.gov.companieshouse.web.accounts.util.ValidationHelper;
 
 @SpringBootApplication
 public class CompanyAccountsWebApplication implements WebMvcConfigurer {
 
     public static final String APPLICATION_NAME_SPACE = "company-accounts.web.ch.gov.uk";
+    public static final String APPLICATION_MODEL_PACKAGE = "uk.gov.companieshouse.web.accounts.model";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(APPLICATION_NAME_SPACE);
 
     private UserDetailsInterceptor userDetailsInterceptor;
     private LoggingInterceptor loggingInterceptor;
@@ -25,6 +32,15 @@ public class CompanyAccountsWebApplication implements WebMvcConfigurer {
     }
 
     public static void main(String[] args) {
+
+        try {
+            ClassPathScanningCandidateComponentProvider provider = new ClassPathScanningCandidateComponentProvider(true);
+            ValidationHelper.scanPackageForValidationMappings(provider, APPLICATION_MODEL_PACKAGE);
+        } catch (RuntimeException e) {
+            LOGGER.error(e.getMessage());
+            System.exit(1);
+        }
+
         SpringApplication.run(CompanyAccountsWebApplication.class, args);
     }
 
--- a/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
@@ -10,7 +10,7 @@ import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 import uk.gov.companieshouse.web.accounts.interceptor.LoggingInterceptor;
 import uk.gov.companieshouse.web.accounts.interceptor.UserDetailsInterceptor;
-import uk.gov.companieshouse.web.accounts.util.ValidationHelper;
+import uk.gov.companieshouse.web.accounts.util.ValidationContext;
 
 @SpringBootApplication
 public class CompanyAccountsWebApplication implements WebMvcConfigurer {
@@ -35,7 +35,7 @@ public class CompanyAccountsWebApplication implements WebMvcConfigurer {
 
         try {
             ClassPathScanningCandidateComponentProvider provider = new ClassPathScanningCandidateComponentProvider(true);
-            ValidationHelper.scanPackageForValidationMappings(provider, APPLICATION_MODEL_PACKAGE);
+            ValidationContext.scanPackageForValidationMappings(provider, APPLICATION_MODEL_PACKAGE);
         } catch (RuntimeException e) {
             LOGGER.error(e.getMessage());
             System.exit(1);
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/BaseController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/BaseController.java
@@ -2,6 +2,7 @@ package uk.gov.companieshouse.web.accounts.controller;
 
 import org.springframework.ui.Model;
 import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.ModelAttribute;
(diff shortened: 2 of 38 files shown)


src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,233 @@
...
+        mockComponentScanning(RECURSIVE_BEAN_NAME);
+        assertThrows(IllegalStateException.class, () -> ValidationHelper.scanPackageForValidationMappings(mockProvider, BASE_PATH));
+    }
+
+    @Test
+    public void testGetValidationError() throws JsonProcessingException {
+        mockComponentScanning(BEAN_CLASS_NAME);
+        ValidationHelper.scanPackageForValidationMappings(mockProvider, BASE_PATH);
+        
+        List<ApiError> errors = createErrors(1);
```
BSaunders95: Don't need to assign a variable here, just use `createErrors(1)` in the method call below

src/main/java/uk/gov/companieshouse/web/accounts/util/ValidationHelper.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Returns a set comprising the class objects representing the primitive
+     * types and String class.
+     *
+     * @return a set of class objects
+     */
+    private static Set<Class<?>> createPrimitivesSet() {
+        Set<Class<?>> types = new HashSet<>();
```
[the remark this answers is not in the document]
marcransome: A very good point; the intention here is to exclude any types we know will not be linked to an API validation error and so save time during the initial validation scan on startup. I'll take a look at what options we have for excluding collection types.

src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,233 @@
...
+        assertEquals(WEB_ERROR_MESSAGE, validationError.getMessageKey());
+    }
+    
+    @Test
+    public void testGetMultipleValidationErrors() throws JsonProcessingException {
+        mockComponentScanning(BEAN_CLASS_NAME);
+        ValidationHelper.scanPackageForValidationMappings(mockProvider, BASE_PATH);
+        
+        int count = 2;
+        List<ApiError> errors = createErrors(count);
```
BSaunders95: As above

src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,233 @@
...
+        assertNotNull(validationErrors);
+        assertEquals(1, validationErrors.size());
+        
+        ValidationError validationError = validationErrors.get(0);
+        assertEquals(FIELD_PATH, validationError.getFieldPath());
+        assertEquals(WEB_ERROR_MESSAGE, validationError.getMessageKey());
+    }
+    
+    @Test
+    public void testGetMultipleValidationErrors() throws JsonProcessingException {
```
BSaunders95: This test is essentially a more generic case of the one above, so I don't believe the previous test is required.

src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,233 @@
...
+        
+        int count = 2;
+        List<ApiError> errors = createErrors(count);
+        ApiErrorResponseException exception = createApiErrorResponse(errors);
+        
+        List<ValidationError> validationErrors = helper.getValidationErrors(exception);
+        assertNotNull(validationErrors);
+        assertEquals(count, validationErrors.size());
+        
+        for(ValidationError validationError : validationErrors) {
```
BSaunders95: Could use a stream here, not that it's too much more performant for a low number of entries in a set. This is preference so it's entirely up to you

src/main/java/uk/gov/companieshouse/web/accounts/controller/BaseController.java
```diff
@@ -19,4 +27,44 @@ protected void addBackPageAttributeToModel(Model model, String... pathVars) {
...
+    /**
+     * Binds one or more API validation errors to model object fields.
+     *
+     * @param bindingResult the binding result object
+     * @param errors        the list of validation errors generated from an
+     *                      API request
+     */
+    protected void bindValidationErrors(BindingResult bindingResult, List<ValidationError> errors) {
+        Collections.sort(errors, Comparator.comparing(ValidationError::getFieldPath).thenComparing(ValidationError::getMessageKey));
+        errors.forEach(error -> {
```
ChrisJBurns: don't need the open braces here

src/main/java/uk/gov/companieshouse/web/accounts/controller/BaseController.java
```diff
@@ -19,4 +27,44 @@ protected void addBackPageAttributeToModel(Model model, String... pathVars) {
...
+     * @param errorArgs the map of error arguments
+     * @return          an array of message argument values
+     */
+    private static Object[] getValidationArgs(Map<String, String> errorArgs) {
+
+        if (errorArgs == null) {
+            return new Object[]{};
+        }
+
+        if (errorArgs.get("lower") != null && errorArgs.get("upper") != null) {
```
BSaunders95: This could be unit tested; just checked Sonar and the condition isn't covered

src/main/java/uk/gov/companieshouse/web/accounts/util/ValidationHelper.java
```diff
@@ -0,0 +1,200 @@
...
+     * @return     the model object field path for the JSON path error string
+     */
+    private static String getModelPathForErrorPath(String path) {
+
+        if (mappings == null) {
+            throw new MissingValidationMappingException("No validation mappings found or validation scanning has not been performed");
+        }
+
+        String modelPath = mappings.get(path);
+        if (modelPath == null || modelPath.isEmpty()) {
```
ChrisJBurns: Can do `StringUtils.isEmpty(modelPath)` here.

src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+    
+    /**
+     * Create a valid API exception with dummy data.
+     * 
+     * @param errors
+     * @return exception
+     * @throws JsonProcessingException
+     */
+    private ApiErrorResponseException createApiErrorResponse(List<ApiError> errors) throws JsonProcessingException {
```
ChrisJBurns: `JsonProcessingException` isn't thrown anywhere in this test so can remove

src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
ChrisJBurns: It would actually be more performant to do a stream here

src/main/java/uk/gov/companieshouse/web/accounts/util/ValidationHelper.java
```diff
@@ -0,0 +1,200 @@
...
+     * @param basePackage       the base package from which to begin scanning
+     * @see   ValidationModel
+     * @see   ValidationMapping
+     */
+    public static void scanPackageForValidationMappings(ClassPathScanningCandidateComponentProvider scanner, String basePackage) {
+
+        scanner.addIncludeFilter(new AnnotationTypeFilter(ValidationModel.class));
+
+        mappings = new HashMap<>();
+        depth = 0;
```
ChrisJBurns: is this thread safe?

src/main/java/uk/gov/companieshouse/web/accounts/util/ValidationHelper.java
```diff
@@ -0,0 +1,200 @@
...
+     * Returns a list of validation errors extracted from the
+     * {@code ApiErrorResponseException} object.
+     *
+     * @param  apiException the exception object from which to
+     *                      extract validation errors
+     * @return              a list of validation errors
+     */
+    public List<ValidationError> getValidationErrors(ApiErrorResponseException apiException) {
+
+        List<ApiError> apiErrors = apiException.getDetails().getErrors();
```
ChrisJBurns: to be honest, you don't really need to assign here can just do:
```java
    public List<ValidationError> getValidationErrors(ApiErrorResponseException apiException) { 
        return apiException.getDetails().getErrors()
                .stream()
                .map(apiError -> {
                    ValidationError validationError = new ValidationError();
                    validationError.setFieldPath(ValidationHelper.getModelPathForErrorPath(apiError.getLocation()));
                    validationError.setMessageKey(ValidationMessage.getMessageKeyForApiError(apiError.getError()));
                    validationError.setMessageArguments(apiError.getErrorValues());
                    return validationError;
        }).collect(Collectors.toList());
```

src/main/java/uk/gov/companieshouse/web/accounts/controller/BaseController.java
```diff
@@ -19,4 +27,44 @@ protected void addBackPageAttributeToModel(Model model, String... pathVars) {
...
     }
+
+    /**
+     * Binds one or more API validation errors to model object fields.
+     *
+     * @param bindingResult the binding result object
+     * @param errors        the list of validation errors generated from an
+     *                      API request
+     */
+    protected void bindValidationErrors(BindingResult bindingResult, List<ValidationError> errors) {
```
ch-mcabrera: Could we format this file as some of the lines are a bit too long? and I have noticed that happens in other files.

src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
```diff
@@ -25,6 +32,15 @@ public CompanyAccountsWebApplication(UserDetailsInterceptor userDetailsIntercept
     }
 
     public static void main(String[] args) {
+
+        try {
+            ClassPathScanningCandidateComponentProvider provider = new ClassPathScanningCandidateComponentProvider(true);
+            ValidationHelper.scanPackageForValidationMappings(provider, APPLICATION_MODEL_PACKAGE);
```
jterrych: This is a Spring component and shouldnt need and static methods to make it to work, the act of spring injecting the bean should be enough for it to do everything thats required. perhaps the static methods should be instance methods instead and the scan scanPackageForValidationMappings be called witihn the constructor?

src/test/java/uk/gov/companieshouse/web/accounts/util/ValidationHelperTest.java
marcransome: Based on a 3-execution comparison between a procedural and functional algorithm here (with results averaged and rounded up to two decimal places), the difference in execution time  averaged 0.53 milliseconds in favour of the procedural method used here. Streams aren't a magical performance optimisation, rather they should be used when the context justifies their use.

marcransome: Based on a 3-execution comparison between a procedural and functional algorithm here (with results averaged and rounded up to two decimal places), the difference in execution time  averaged 0.53 milliseconds in favour of the procedural method used here. Streams aren't a magical performance optimisation, rather they should be used when the context justifies their use.

jterrych: parallel streams allow for running the assertions and tests in parallel which will almost certainly be a performance gain, and an improvement going forward as devs learn how to use them. Its as simple as calling parallelStream() or steam().parallel() rather than just stream(). This is not something you can do so easily the procedural way (and you wouldn't want to either).

marcransome: Parallel streams involve a higher overhead than the actual processing effort involved in the loop here. In the case of both `parallelStream()` and `stream().parallel()` both perform worse than the traditional `for` loop.

Given that we're in the context of a unit test here, dealing with a list of fixed-size never likely to change, any performance gain would have negligible benefit, not to mention that cursory testing shows it would actually perform slower. I'm happy to discuss further, but would consider this to be premature optimisation at best.

src/main/java/uk/gov/companieshouse/web/accounts/util/ValidationHelper.java
```diff
@@ -0,0 +1,200 @@
...
+ * The {@code scanPackageForValidationMappings} method must be called at
+ * startup to initialise the list of mappings between the API JSON error
+ * paths and model fields that other methods of this class depend on.
+ *
+ * @see ValidationModel
+ * @see ValidationMapping
+ * @see ValidationMessage
+ */
+@Component
+public class ValidationHelper {
```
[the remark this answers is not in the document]
marcransome: This is a very valid point. Given that the class now encapsulates some persistent state for the lifetime of the app, `context` would seem a better candidate for the class name than the original `helper` suffix. I'll rename accordingly.

src/main/java/uk/gov/companieshouse/web/accounts/util/ValidationHelper.java
marcransome: As @BSaunders95 has suggested, the scan is executed once on application startup from the main thread, so this shouldn't be a concern. In general, there are no guarantees as to the thread safety of methods in this class—if we change the way validation mappings are scanned in future this may need to be considered.

PR #78, opened 2018-10-01
Add piwik to criteria page

ch-mcabrera:
Code changes to add piwik to the criteria page when the radio buttons are clicked.
This change has followed similar changes made in the transaction web. See Pr for further info:  https://github.com/companieshouse/transactions.web.ch.gov.uk/pull/16)

Resolves: SFA-653

### DIFF
--- a/src/main/resources/templates/fragments/piwik.html
+++ b/src/main/resources/templates/fragments/piwik.html
@@ -7,7 +7,7 @@
 <div th:fragment="piwik">
   <script th:inline="javascript">
                 (function() {
-                    bindPiwikListener('web', /*[[${@environment.getProperty('piwik.url')}]]*/, /*[[${@environment.getProperty('piwik.siteId')}]]*/);
+                    bindPiwikListener('company-accounts', /*[[${@environment.getProperty('piwik.url')}]]*/, /*[[${@environment.getProperty('piwik.siteId')}]]*/);
                 })();
             </script>
   <noscript>
--- a/src/main/resources/templates/fragments/piwik.html
+++ b/src/main/resources/templates/fragments/piwik.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+<head>
+  <title>Piwik</title>
+</head>
+<body>
+<div th:fragment="piwik">
+  <script th:inline="javascript">
+                (function() {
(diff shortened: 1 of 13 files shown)


src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
```diff
@@ -68,6 +68,8 @@
 
     private static final String BACK_BUTTON_MODEL_ATTR = "backButton";
 
+    private static final String TEMPLATE_NAME_MODEL_ATTR = "templateName";
```
BSaunders95: Can you assert this attribute exists when the page is redisplayed after an error?

src/main/resources/templates/fragments/piwik.html
```diff
@@ -0,0 +1,21 @@
...
+<html xmlns:th="http://www.thymeleaf.org">
+<head>
+  <title>Piwik</title>
+</head>
+<body>
+<div th:fragment="piwik">
+  <script th:inline="javascript">
+                (function() {
+                    bindPiwikListener('web', /*[[${@environment.getProperty('piwik.url')}]]*/, /*[[${@environment.getProperty('piwik.siteId')}]]*/);
+                })();
```
marcransome: Is there a reason the function call arguments are commented out here? This isn't valid JavaScript: `Uncaught SyntaxError: Unexpected token `.

src/main/resources/templates/fragments/piwik.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+<head>
+  <title>Piwik</title>
+</head>
+<body>
+<div th:fragment="piwik">
+  <script th:inline="javascript">
+                (function() {
+                    bindPiwikListener('web', /*[[${@environment.getProperty('piwik.url')}]]*/, /*[[${@environment.getProperty('piwik.siteId')}]]*/);
```
BSaunders95: Can the first argument here be changed to 'company-accounts' for the clarity in the Piwik dashboard?

BSaunders95:
As part of these changes, a separate PR must be raised for the `enable-piwik.js` script to remove ids from the url paths. In the `setCustomUrls` method, we should replace company number with `_number`

src/main/resources/templates/fragments/piwik.html
ch-mcabrera: Perfect, I will do. I wasn't sure what the argument i had to used

PR #81, opened 2018-10-02
Update SDK version to the latest major release

BSaunders95:
Update SDK version to the latest major release, and change model integers to longs to prevent type castings exceptions when overflowing an integer's size.

Resolves SFA-791 and SFA-790

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
@@ -2,8 +2,11 @@ package uk.gov.companieshouse.web.accounts.service.transaction.impl;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.web.util.UriTemplate;
 import uk.gov.companieshouse.api.ApiClient;
 import uk.gov.companieshouse.api.error.ApiErrorResponseException;
+import uk.gov.companieshouse.api.handler.exception.URIValidationException;
+import uk.gov.companieshouse.api.handler.transaction.TransactionsResourceHandler;
 import uk.gov.companieshouse.api.model.transaction.Transaction;
 import uk.gov.companieshouse.api.model.transaction.TransactionStatus;
 import uk.gov.companieshouse.web.accounts.api.ApiClientService;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
```diff
@@ -2,8 +2,11 @@
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.web.util.UriTemplate;
 import uk.gov.companieshouse.api.ApiClient;
 import uk.gov.companieshouse.api.error.ApiErrorResponseException;
+import uk.gov.companieshouse.api.handler.exception.URIValidationException;
+import uk.gov.companieshouse.api.handler.transaction.TransactionsResourceHandler;
```
marcransome: Is this import required?

src/test/java/uk/gov/companieshouse/web/accounts/service/company/impl/CompanyServiceImplTests.java
```diff
@@ -33,24 +35,33 @@
...
+    private CompanyGet companyGet;
+
     @InjectMocks
     private CompanyService companyService = new CompanyServiceImpl();
 
     private static final String COMPANY_NUMBER = "companyNumber";
 
+    private static final String COMPANY_URI = "/company/" + COMPANY_NUMBER;
+
+    private static final String INVALID_COMPANY_URI = COMPANY_URI + "/blah";
```
marcransome: Humour aside—and it did give me a chuckle—we should be aiming to keep the codebase as professional as possible, perhaps `invalid-company-uri` or similar would work better here rather than `blah`.

marcransome:
The additional changes look good. The only remaining issue is the failing build which may require updating the `pom.xml` file to account for the move to trunk-based development of the `api-sdk-java` dependency.

src/test/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformerTests.java
```diff
@@ -13,7 +13,7 @@
 
 public class BalanceSheetTransformerTests {
 
-    private Integer CALLED_UP_SHARE_CAPITAL = 1;
+    private Long CALLED_UP_SHARE_CAPITAL = Long.valueOf("1");
```
marcransome: For reference, this could be written as `private Long CALLED_UP_SHARE_CAPITAL = 1L;`.

PR #87, opened 2018-10-08
Adds previous period called up share capital not paid resource to balance sheet

ChrisJBurns:
- adds previous amounts to `CalledUpSharedCapitalNotPaid` model
- adds previous period logic to balance sheet service
- adds previous period called up share capital not paid value on balance sheet template
- refactors balance sheet service tests so they are more structured
- adds link types to prevent duplicating string literals across the microservice

Resolves: SFA-751


src/main/java/uk/gov/companieshouse/web/accounts/links/SmallFullLinkType.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.companieshouse.web.accounts.links;
+
+public enum SmallFullLinkType implements LinkType {
```
BSaunders95: Do we need to implement an interface here? Is it just for the purpose of unit testing?

ChrisJBurns: Nope, just copying what was done in the api :) .. It's handy to have enums as it stops the amount of string literals pasted everywhere in the code further making it easier to change - if a link ever gets changed in the future

BSaunders95:
The build is currently failing, just a heads up http://ch-jenkins-2k8.internal.ch:8080/jenkins/job/Build/job/chs/job/company-accounts.web.ch.gov.uk/job/feature/367/

src/main/java/uk/gov/companieshouse/web/accounts/links/SmallFullLinkType.java
BSaunders95: Fair enough. If that's the case, no need for the `LinkType` interface if the `getLink` method is only implemented here.

ChrisJBurns:
Yep, I'm aware, it is the unit tests for the PUT tests. These shall be passing when the API has been done for the PUT. Bare in mind, it will NOT be merged until it all passes. I have written them following TDD, and the tests shall pass the minute the PUT resource is in the API :+1:

src/main/java/uk/gov/companieshouse/web/accounts/links/SmallFullLinkType.java
ChrisJBurns: I think it's best to leave the `LinkType` in as eventually it would be worth adding `CompanyAccountLinkType` etc to extend from it - again following the `company-accounts.api`

BSaunders95:
Cool by me, just wanted to raise it incase you weren't aware 👍

src/main/java/uk/gov/companieshouse/web/accounts/links/SmallFullLinkType.java
BSaunders95: Okay yep that makes sense to me. Disregard that then.

ChrisJBurns:
As discussed, I have inserted the PUT lines in just for the tests to pass and to aid in the review process so you know what is going in. I believe the branch on Jenkins is building correctly and all tests are passing. So that should make this PR fully ready for review. :+1:

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
+                apiClient.smallFull().previousPeriod().update(previousPeriodUri, previousPeriodApi).execute();
+            }
 
+        } catch (ApiErrorResponseException e) {
             if (e.getStatusCode() == HttpStatus.BAD_REQUEST.value()) {
-                List<ValidationError> validationErrors = validationContext.getValidationErrors(e);
-                if (validationErrors == null) {
+                validationErrors.addAll(validationContext.getValidationErrors(e));
+                if (validationErrors.isEmpty()) {
                     throw new ServiceException("Bad request posting balance sheet", e);
```
marcransome: As this exception is now thrown in the context of the `createPreviousPeriod()` method, it should be given a more specific exception message to differentiate from a similar exception thrown in the `createCurrentPeriod`() method, for example: `Bad request when creating previous period resource`.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
+            } else {
+                apiClient.smallFull().currentPeriod().update(currentPeriodUri, currentPeriod).execute();
             }
 
-            throw new ServiceException("Error posting balance sheet", e);
+        } catch (ApiErrorResponseException e) {
+            if (e.getStatusCode() == HttpStatus.BAD_REQUEST.value()) {
+                validationErrors.addAll(validationContext.getValidationErrors(e));
+                if (validationErrors.isEmpty()) {
+                    throw new ServiceException("Bad request posting balance sheet", e);
```
marcransome: As above, the exception message here should be made more specific to differentiate it from a similar exception thrown in `createPreviousPeriod()`, for example: `Bad request when creating current period resource`.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
+    }
+
+    private boolean isMultipleYearFiler(CompanyProfileApi companyProfile) {
+        LastAccountsApi lastAccountsApi = companyProfile.getAccounts().getLastAccounts();
+        return lastAccountsApi != null && lastAccountsApi.getPeriodEndOn() != null;
+    }
 
-        Date date = new Date(dateTime.getValue());
-        return accountsDatesHelper.convertDateToLocalDate(date);
+    private boolean isCurrentPeriodCreated(Map<String, String> links) {
```
marcransome: This method would be better named using the `has` prefix convention, e.g. `hasCurrentPeriod()`.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
+        return isCreated(links, SmallFullLinkType.CURRENT_PERIOD);
+    }
+
+    private boolean isPreviousPeriodCreated(Map<String, String> links) {
```
marcransome: Similarly, this method would also be better named using the `has` prefix convention, e.g. `hasPreviousPeriod()`.

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
```diff
@@ -8,4 +9,11 @@
     BalanceSheet getBalanceSheet(CurrentPeriodApi currentPeriod);
 
     CurrentPeriodApi getCurrentPeriod(BalanceSheet balanceSheet);
+
+    /**
+     * Gets the previous period data on the balance sheet
+     * @param balanceSheet populated balance sheet
```
marcransome: For consistency with our other Javadoc comments and general convention, could you add a linebreak before the `@param..` line and end the description itself with a full-stop?

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
-            LocalDate previousPeriodEndOn =
-                convertDateTimeToLocalDate(companyProfile.getAccounts().getLastAccounts().getPeriodEndOn());
-            isSameYear = accountsDatesHelper.isSameYear(previousPeriodEndOn, currentPeriodEndOn);
+            return accountsDatesHelper.generateBalanceSheetHeading(previousPeriodStartOn, previousPeriodEndOn, isSameYear);
         }
+        return null;
+    }
 
-        String currentPeriodHeading = accountsDatesHelper.generateBalanceSheetHeading(currentPeriodStartOn, currentPeriodEndOn, isSameYear);
+    private boolean isSameYear(CompanyProfileApi companyProfile) {
```
marcransome: The `isSameYear()` method would be better named `isSameYearFiler()` to add additional context, and for consistency with the `isMultipleYearFiler()` method.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
+        }
+        return false;
     }
 
-    private LocalDate convertDateTimeToLocalDate(DateTime dateTime) {
+    private CompanyProfileApi getCompanyProfile(String companyNumber) throws ServiceException {
+        return companyService.getCompanyProfile(companyNumber);
+    }
+
+    private String createUri(UriTemplate uri, String transactionId, String companyAccountsId) {
```
BSaunders95: Is this method required? Since we're returning a single method call it may read more simply to just call the uri `expand` method inline

marcransome: I'd add to that by saying we'll be using `expand()` in many other places, and having a custom convenience method just to call `toString()` on the result feels unnecessary and will end up deviating from what we do in other places, reducing overall consistency. There's also the risk that `createUri()` gets duplicated in other services implementations.

src/main/resources/templates/fragments/fieldErrorRow.html
```diff
@@ -14,10 +14,16 @@
...
                 <div class="govuk-grid-column-one-quarter">
-                    <!-- This fragment will be expanded to support a previous field -->
-                    &nbsp;
+                    <span th:if="${#fields.hasErrors('__${previousField}__')}"
+                          th:id="${previousField} + '-errorId'"
+                          th:each="e : ${#fields.errors('__${previousField}__')}"
+                          th:text="${e}"
+                          class="govuk-error-message">
+                    </span>
+
```
BSaunders95: Do we need an empty span here if there are no field errors for the previous field?

src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
```diff
@@ -87,341 +114,619 @@
...
+        return nextAccounts;
+    }
 
-        LastAccountsApi lastAccounts = new LastAccountsApi();
-        lastAccounts.setPeriodEndOn(previousPeriodEnd);
-
-        CompanyAccountApi companyAccounts = new CompanyAccountApi();
-        companyAccounts.setNextAccounts(nextAccounts);
-        companyAccounts.setLastAccounts(lastAccounts);
+    private void mockApiClientGet() {
```
BSaunders95: This method may be better renamed as `mockApiClientGetSmallFull` for clarity

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
ChrisJBurns: Since this exception can be thrown on the `POST` and `PUT` I will instead say, `Bad request when submitting previous period resource` as I feel `creating` gives the impression that it is just a `POST` and not `PUT`

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
ChrisJBurns: I added it just because I was following the DRY principle, but if you feel it isn't needed then I will remove.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -72,66 +83,152 @@ public BalanceSheet getBalanceSheet(String transactionId, String companyAccounts
...
-        return accountsDatesHelper.convertDateToLocalDate(date);
+    private boolean isCurrentPeriodCreated(Map<String, String> links) {
+        return isCreated(links, SmallFullLinkType.CURRENT_PERIOD);
+    }
+
+    private boolean isPreviousPeriodCreated(Map<String, String> links) {
+        return isCreated(links, SmallFullLinkType.PREVIOUS_PERIOD);
+    }
+
+    private boolean isCreated(Map<String, String> links, SmallFullLinkType linkType) {
```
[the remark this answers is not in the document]
ChrisJBurns: The reason it was done this way was because I felt instead of having a `isPreviousPeriodCreated` method that excepts 2 parameters which can increase the line length, it may be easier to just have a method that does it all for you.

Regarding the class name, I have followed what is done in the API and I didn't want to change because then we are out of sync with naming. So I feel the same change would have to be made there as well to keep things consistent

src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
ChrisJBurns: I'm not sure about that as we are in the small full parent package and that would mean I would have to add small full onto every method and test to keep consistency for that same rule. Welcome to be overruled if others feel the same but I do feel that because we are in the small full package that it is to be assumed that it will only be doing small full related functionality

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformer.java
ChrisJBurns: resolved

src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImplTests.java
```diff
@@ -411,6 +455,29 @@ void postMultipleYearFilerBalanceSheetCurrentPeriodFailureBadRequestWithNoValida
...
+    void postMultipleYearFilerBalanceSheetCurrentPeriodFailureNonBadRequestWithNoValidationErrors() throws ApiErrorResponseException, URIValidationException, ServiceException {
+        mockApiClient();
+        createMultipleYearFilerCompanyProfile();
+        createMultipleYearFilerSmallFullAccountPost();
+
+        BalanceSheet balanceSheet = createMultipleYearFilerBalanceSheetTestObject();
+        mockPreviousPeriodPost(balanceSheet);
+        mockCurrentPeriodPost(balanceSheet);
+
+        HttpResponseException httpResponseException = new HttpResponseException.Builder(404,"Bad Request",new HttpHeaders()).build();
```
BSaunders95: Should this not be a 404? Same with the others, too

ChrisJBurns: Well, it can be anything, its just something other than `400` as that is what we are testing.

BSaunders95: Ah sorry, just being an idiot 👍

PR #91, opened 2018-10-15
Create review page

marcransome:
The changes here introduce the following:

* Add controller, service, model, template, and template fragments for review page
* Add unit tests for service and controller
* Update balance sheet unit tests


src/main/java/uk/gov/companieshouse/web/accounts/util/Navigator.java
```diff
@@ -35,7 +35,7 @@ public static String getNextControllerRedirect(Class clazz, String... pathVars)
             throw new MissingAnnotationException("Missing @RequestMapping value on class " + nextControllerClass.toString());
         }
 
-        return UrlBasedViewResolver.REDIRECT_URL_PREFIX + new UriTemplate(mappings[0]).expand(pathVars);
+        return UrlBasedViewResolver.REDIRECT_URL_PREFIX + new UriTemplate(mappings[0]).expand((Object[]) pathVars);
```
BSaunders95: Do we need the `Object[]` cast here?

src/main/resources/templates/fragments/review/sectionHeader.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org">
+    <head>
+        <title>Section Header</title>
+    </head>
+    <body>
+        <h2 th:fragment="sectionHeader" class="govuk-heading-m"><th:block th:text="${description}"></th:block>
```
BSaunders95: Are we going to use this fragment on other pages? If not, is it worth renaming to be review specific?

src/main/resources/templates/smallfull/review.html
```diff
@@ -0,0 +1,85 @@
...
+                        <dd class="app-check-your-answers__answer govuk-heading-s"
+                            th:if="*{balanceSheet.balanceSheetHeadings.previousPeriodHeading != null}"
+                            th:text='*{balanceSheet.balanceSheetHeadings.previousPeriodHeading}'></dd>
+                        <dd class="app-check-your-answers__change"></dd>
+                    </div>
+
+                    <div th:replace="fragments/review/balanceSheetRow :: balanceSheetRow (
+                        description = 'Called up share capital not paid',
+                        identifier = 'called-up-share-capital-not-paid',
+                        currentAmount = *{balanceSheet?.calledUpShareCapitalNotPaid?.currentAmount},
```
BSaunders95: Shouldn't need the '?' operator following balanceSheet, not that it makes much difference

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
```diff
@@ -63,6 +61,11 @@
                                                 "/company-accounts/" + COMPANY_ACCOUNTS_ID +
                                                 "/small-full/approval";
 
+    private static final String REVIEW_PATH = "/company/" + COMPANY_NUMBER +
```
BSaunders95: The `APPROVAL_PATH` above should be redundant now and can be removed

src/main/resources/templates/smallfull/review.html
marcransome: I agree, however I think the single `?` buried in the middle of this string makes it harder to discern—this is only required where a nested field may be `null` (and only for optional fields on the balance sheet such as the called up share capital). I would prefer to use this for all fields that may be null in a string, or none, depending on whether the field is mandatory or optional.

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetControllerTests.java
marcransome: Removed. 👍

src/main/resources/templates/fragments/review/sectionHeader.html
marcransome: I've added a `review` subdirectory to `fragments` rather than pollute the name of each fragment file. If there's a preference for filename prefixes or suffixes I can change this?

src/main/resources/templates/fragments/review/balanceSheetRow.html
```diff
@@ -0,0 +1,23 @@
...
+                th:classappend="${fontBold} ? 'govuk-!-font-weight-bold' : ''">
+                <span class="mobile-only-label" th:text="*{'(' + balanceSheet.balanceSheetHeadings.currentPeriodHeading + ')'}"></span>
+            </dd>
+            <dd th:if="*{balanceSheet.balanceSheetHeadings.previousPeriodHeading != null}"
+                th:attr="id=${identifier != null} ? 'review-' + ${identifier} + '-previous-amount' : null"
+                th:text="${previousAmount != null ? '£' + #numbers.formatInteger(previousAmount, 0, 'COMMA') : '£0'}" class="app-check-your-answers__answer"
+                th:classappend="${fontBold} ? 'govuk-!-font-weight-bold' : ''">
+                <span class="mobile-only-label" th:text="*{'(' + balanceSheet.balanceSheetHeadings.currentPeriodHeading + ')'}"></span>
+            </dd>
+            <dd class="app-check-your-answers__change"></dd>
```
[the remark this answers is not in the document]
marcransome: There are media queries and other styling applied to the class associated with this element—which appears in our prototype—I would be hesitant to remove without further investigation and discussion surrounding the latest iteration of the prototype.

src/main/java/uk/gov/companieshouse/web/accounts/util/Navigator.java
marcransome: This was added to silence IDE warnings regarding ambiguity in this method call—we could modify the method signature to accept a variable argument parameter of type `Object...` but this might obscure the actual underlying `String` type expected. For the sake of silencing the warnings, I think this may be the better option.

PR #97, opened 2018-10-23
Add current assets web functionality and validation

leahtaylor:
Update sdk version in POM
Add @ValidationMapping annotations to current assets fields
Add current assets to the transformer and transformer unit tests
Add field error fragments to current assets fields on templates
Add current assets fields to message.properties for specific error messages.

SFA-205, SFA-459

*base branch will change when feature/add-current-assets is merged
* tech debt - validation to be moved from controller to service layer

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CashAtBankAndInHand.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CashAtBankAndInHand.java
@@ -1,0 +1,13 @@
+package uk.gov.companieshouse.web.accounts.model.smallfull;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class CashAtBankAndInHand {
+
+    private Long currentAmount;
+    private Long previousAmount;
+
+}
--- a/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CashAtBankAndInHand.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CashAtBankAndInHand.java
@@ -2,12 +2,16 @@ package uk.gov.companieshouse.web.accounts.model.smallfull;
 
 import lombok.Getter;
 import lombok.Setter;
+import uk.gov.companieshouse.web.accounts.validation.ValidationMapping;
 
 @Getter
 @Setter
 public class CashAtBankAndInHand {
 
+    @ValidationMapping("$.current_period.balance_sheet.current_assets.cash_at_bank_and_in_hand")
     private Long currentAmount;
+
+    @ValidationMapping("$.previous_period.balance_sheet.current_assets.cash_at_bank_and_in_hand")
     private Long previousAmount;
 
 }
--- a/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
@@ -1,8 +1,19 @@
 package uk.gov.companieshouse.web.accounts.transformer.smallfull.impl;
 
 import org.springframework.stereotype.Component;
-import uk.gov.companieshouse.api.model.accounts.smallfull.*;
-import uk.gov.companieshouse.web.accounts.model.smallfull.*;
+import uk.gov.companieshouse.api.model.accounts.smallfull.BalanceSheetApi;
+import uk.gov.companieshouse.api.model.accounts.smallfull.CurrentPeriodApi;
+import uk.gov.companieshouse.api.model.accounts.smallfull.PreviousPeriodApi;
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
```diff
@@ -142,6 +211,63 @@ private void populatePreviousFixedAssets(BalanceSheet balanceSheet, FixedAssetsA
...
+            cashAtBankAndInHand.setPreviousAmount(currentAssetsApi.getCashAtBankAndInHand());
+        }
+
+        // Total
+        if (currentAssetsApi.getTotal() != null) {
+            currentAssets.setPreviousCurrentAssetsTotal(currentAssetsApi.getTotal());
+        }
+    }
+
+    private void populateCurrentCurrentAssets(BalanceSheet balanceSheet, CurrentAssetsApi currentAssetsApi) {
```
BSaunders95: Could we maybe rename this as something like `populateCurrentPeriodCurrentAssets` for clarity? If we chose to do this, it would also be worth renaming the other methods to make them consistent

src/test/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformerTests.java
```diff
@@ -2,14 +2,8 @@
...
 import org.junit.jupiter.api.Test;
-import uk.gov.companieshouse.api.model.accounts.smallfull.BalanceSheetApi;
-import uk.gov.companieshouse.api.model.accounts.smallfull.CurrentPeriodApi;
-import uk.gov.companieshouse.api.model.accounts.smallfull.FixedAssetsApi;
-import uk.gov.companieshouse.api.model.accounts.smallfull.PreviousPeriodApi;
-import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
-import uk.gov.companieshouse.web.accounts.model.smallfull.CalledUpShareCapitalNotPaid;
-import uk.gov.companieshouse.web.accounts.model.smallfull.FixedAssets;
-import uk.gov.companieshouse.web.accounts.model.smallfull.TangibleAssets;
+import uk.gov.companieshouse.api.model.accounts.smallfull.*;
```
BSaunders95: Can we keep imports explicit please?

src/test/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/BalanceSheetTransformerTests.java
```diff
@@ -20,10 +14,18 @@
     private static final Long CURRENT_CALLED_UP_SHARE_CAPITAL = 1L;
     private static final Long CURRENT_TANGIBLE = 5L;
     private static final Long CURRENT_FIXED_ASSETS_TOTAL = 5L;
+    private static final Long CURRENT_STOCKS = 6L;
```
BSaunders95: This may be a bit nit-picky, but can we keep these values distinct please? That way we ensure unit tests don't mix up values. Marc used different values, but ones which are real-world (so the total for the section is correct), and then he used twice the value for the previous period

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/BalanceSheetTransformerImpl.java
```diff
@@ -142,6 +211,63 @@ private void populatePreviousFixedAssets(BalanceSheet balanceSheet, FixedAssetsA
...
+        CurrentAssets currentAssets = createCurrentAssets(balanceSheet);
+
+        // Stocks
+        if (currentAssetsApi.getStocks() != null) {
+            Stocks stocks = createStocks(balanceSheet);
+            stocks.setPreviousAmount(currentAssetsApi.getStocks());
+        }
+
+        // Debtors
+        if (currentAssetsApi.getStocks() != null) {
```
ch-mcabrera: I think this should be `currentAssetsApi.getDebtors()`

src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CashAtBankAndInHand.java
```diff
@@ -2,12 +2,16 @@
 
 import lombok.Getter;
 import lombok.Setter;
+import uk.gov.companieshouse.web.accounts.validation.ValidationMapping;
 
 @Getter
 @Setter
 public class CashAtBankAndInHand {
 
+    @ValidationMapping("$.current_period.balance_sheet.current_assets.cash_at_bank_and_in_hand")
```
marcransome: You'll need to add a `@ValidationModel` annotation to this class for these mappings to be picked up.

BSaunders95: I didn't think that was the case? The scanner works recursively from the top level model, and BalanceSheet holds the `@ValidationModel` annotation.

PR #98, opened 2018-10-25
Add current assets review screen

leahtaylor:
Add current assets sections to the balance sheet on the review screen 

Rename current assets total fields to be consistent with other total fields

SFA-533


src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/CurrentAssets.java
```diff
@@ -13,10 +13,9 @@
     private CashAtBankAndInHand cashAtBankAndInHand;
 
     @ValidationMapping("$.current_period.balance_sheet.current_assets.total")
-    private Long currentCurrentAssetsTotal;
+    private Long totalCurrentCurrentAssets;
```
sroberts2: `totalCurrentCurrentAssets` seems excessively long for a field name, does it need to have `CurrentAssets` in the name as its part of the `CurrentAssets` class? Something like `currentTotal` and `previousTotal`. So references to the field would be CurrentAssets.getCurrentTotal.

PR #100, opened 2018-10-29
Add other liabilities and assets fields to balance sheet and review page

marcransome:
This pull-request introduces the following changes:

- Add models and validation mappings to support other liabilities and assets
- Add new section and fields to review page
- Add field description messages for validation errors
- Add balance sheet transforms for other liabilities and assets fields and update unit tests

Resolves: SFA-249, SFA-470, SFA-534


src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/BalanceSheet.java
```diff
@@ -21,4 +21,6 @@
 
     @Valid
     private CurrentAssets currentAssets;
+
+    private OtherLiabilitiesOrAssets otherLiabilitiesOrAssets;
```
BSaunders95: Should this field be annotated with `@Valid`?

src/main/resources/templates/smallfull/balanceSheet.html
```diff
@@ -119,26 +118,29 @@ <h2 class="govuk-heading-m">Fixed assets</h2>
...
                             <strong>Total fixed assets</strong>
                         </div>
-
                         <div class="govuk-grid-column-one-quarter">
                             <div th:replace="fragments/numberInput :: numberInput (
-                                 id = 'fixedAssets.currentTotal',
-                                 text = 'Total current fixed assets',
-                                 class= 'govuk-input current-fixed-assets-total',
-                                 )">
+
```
BSaunders95: Can we keep spacing consistent please?

src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/BalanceSheet.java
marcransome: The `@Valid` annotation would be redundant here as we don't associate any other validation annotations with the field or nested object fields. For consistency I've removed all of the unnecessary annotations.

PR #103, opened 2018-11-06
Add Statements Contoller

BSaunders95:
- Add Statements controller and corresponding tests. 
- Update steps to complete controller to create the statements resource on submission. 
- Update Balance sheet controller to redirect to the statements page on successful submission.

Resolves SFA-50

(Reliant on https://github.com/companieshouse/company-accounts.web.ch.gov.uk/pull/102)


src/main/resources/templates/smallfull/statements.html
```diff
@@ -0,0 +1,44 @@
...
+                        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+                        <strong id="statements-warning-text" class="govuk-warning-text__text">
+                            <span class="govuk-warning-text__assistive">Warning</span>
+                            The following statements about your accounts will be included on the public record.
+                        </strong>
+                    </div>
+
+                    <ul id="statements-list" class="govuk-list govuk-list--bullet">
+                        <li id="statements-section-477"><span th:text="*{section477}"></span></li>
+                        <li id="statements-audit-not-required-by-members"><span th:text="*{auditNotRequiredByMembers}"></span></li>
```
marcransome: Are the `span` elements for each of the list items here required? If they have no effect on the page styling they would seem redundant, in which case the `th:text` attributes can be moved to the `<li>` elements, or alternatively `<th:block>` elements can be used instead to avoid rendering unnecessary elements to the user agent.

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/StepsToCompleteControllerTests.java
```diff
@@ -212,4 +220,35 @@ void postRequestCompanyAccountsServiceCreateSmallFullAccountsFailure() throws Ex
...
+    @Test
+    @DisplayName("Post steps to complete failure path for create statements resource")
+    void postRequestStatementsServiceCreateStatementsFailure() throws Exception {
+
+        when(transactionService.createTransaction(COMPANY_NUMBER)).thenReturn(TRANSACTION_ID);
+
+        LocalDate periodEndOn = LocalDate.now();
+
+        NextAccountsApi nextAccounts = new NextAccountsApi();
+        nextAccounts.setPeriodEndOn(periodEndOn);
```
marcransome: Could we remove the local `periodEndOn` variable above in favour of calling `.setPeriodEndOn(LocalDate.now())`?

PR #119, opened 2018-11-28
Add save and resume support

marcransome:
These changes introduce support for 'save and resume' of the web journey.

**Related pull-requests:**
* https://github.com/companieshouse/abridged.accounts.web.ch.gov.uk/pull/554
* https://github.com/companieshouse/transactions.api.ch.gov.uk/pull/89
* https://github.com/companieshouse/ch.gov.uk/pull/24

Resolves: SFA-47


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/ResumeServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Autowired
+    EnvironmentReader environmentReader;
+
+    private static final String CHS_URL = "CHS_URL";
+
+    @Override
+    public String getResumeRedirect(String companyNumber, String transactionId, String companyAccountsId) {
+
+        return UrlBasedViewResolver.REDIRECT_URL_PREFIX +
```
BSaunders95: Very minor thing, can we move the `UrlBasedViewResolver.REDIRECT_URL_PREFIX` to the controller layer? I believe redirection should be the concern of the controller rather than the service.

marcransome: I'm leaning towards leaving this as-is; the controller is still responsible for directing the flow by performing the redirect from the handler method, but the service itself happens to contain the method that will generate the redirection string.

leahtaylor:
@BSaunders95  conflicts in pom resolved

PR #137, opened 2018-12-17
Correct api error returned when entering an invalid character

leahtaylor:
Correct api error returned when an invalid character is entered

invalid_characters_entered to invalid_character.

Corresponding api change in https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/159

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/enumeration/ValidationMessage.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/enumeration/ValidationMessage.java
@@ -12,7 +12,6 @@ public enum ValidationMessage {
 
     // API validation error to message key mappings
     VALUE_OUTSIDE_RANGE("value_outside_range", "validation.range.outside", true),
-    INVALID_CHARACTER("invalid_character", "validation.character.invalid", true),
     INVALID_INPUT_LENGTH("invalid_input_length", "validation.length.invalidInputLength", false),
     INCORRECT_TOTAL("incorrect_total", "validation.total.invalid", true),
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/enumeration/ValidationMessage.java
```diff
@@ -15,7 +15,7 @@
     INVALID_CHARACTER("invalid_character", "validation.character.invalid", true),
     INVALID_INPUT_LENGTH("invalid_input_length", "validation.length.invalidInputLength", false),
     INCORRECT_TOTAL("incorrect_total", "validation.total.invalid", true),
-    INVALID_CHARACTERS_ENTERED("invalid_characters_entered", "validation.characters.invalid", false),
+    INVALID_CHARACTERS_ENTERED("invalid_character", "validation.characters.invalid", false),
```
BSaunders95: This enum already exists at line 15

src/main/resources/messages.properties
```diff
@@ -61,7 +61,7 @@ date = Approval date:
 
 # Validation
 validation.range.outside = Enter a value between {0} and {1}
-validation.character.invalid = Enter a value without commas, brackets or other symbols
+validation.characters.invalid = Enter a value without commas, brackets or other symbols
```
BSaunders95: I'm not sure this is needed anymore. The 'false' in the enum above indicates that this error is not generic. As a result, the message resolver will look for a message of the format `validation.characters.invalid.<json_path_of_field_in_error>`.

PR #147, opened 2019-01-15
Add conditional navigation support

marcransome:
This pull-request introduces conditional navigation support to the linear journey. This is implemented using a `ConditionalController` interface that defines a `willRender()` method. Any conditional controller should implement this interface and return boolean `true` or `false` to indicate whether the template should be rendered based on the current state of the data.

**Resolves:** SFA-1050


src/main/java/uk/gov/companieshouse/web/accounts/controller/ConditionalController.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.companieshouse.web.accounts.controller;
+
+/**
+ * The {@code ConditionalController} interface defines a single method that
+ * should be implemented in controller classes whose template is rendered
+ * conditionally (i.e. the controller will only render its template
+ * dependent on some condition, such as the preseence or absence of data).
```
ljgeorge: I think 'preseence' might be 'presence'

PR #152, opened 2019-01-17
Add 10 year expiration to cookie so it "never" expires

drosser:
Cookies need to have some kind of expiry, and given the nature of the cookie it needs to last as long as possible. 10 years is *enough* time that a person will probably either change their computer or clear their cookies before then, so it's considered a "never-expiring" cookie.

Resolves: SFA-1051.


src/main/java/uk/gov/companieshouse/web/accounts/interceptor/CompanyAccountsDataStateInterceptor.java
```diff
@@ -127,6 +127,9 @@ public void postHandle(HttpServletRequest request, HttpServletResponse response,
                 stateCookie.setValue(token);
                 stateCookie.setPath("/");
 
+                //Set expiration of cookie to 10 years in the future
+                stateCookie.setMaxAge(60 * 60 * 24 * 365 * 10);
```
ch-mcabrera: we could have this in a constant with a descriptive name ...what do you think?

PR #154, opened 2019-01-18
Move navigation to service layer and add support for error handling

marcransome:
The changes in this pull-request were introduced to avoid an edge case scenario with navigation explained below:

With several adjacent conditional controllers in the controller chain, if the first call to a controller's `willRender()` method results in a request being made to the API and this request **fails**, the method returns `false` meaning the controller will be skipped in the journey. If any subsequent call to `willRender()` succeeds, the associated template is rendered (or alternatively the next mandatory controller in the chain will be rendered). This could lead to pages being skipped during the web journey if requests to the API fail or if the API itself is unavailable.

With these changes, the `willRender()` method is now expected to throw a `ServiceException` for any failed requests, which `NavigatorService` will catch and handle by throwing a custom `RuntimeException` leading to the default error page being returned to the user agent.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/exception/NavigationException.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/exception/NavigationException.java
@@ -1,0 +1,23 @@
+package uk.gov.companieshouse.web.accounts.exception;
+
+/**
+ * The class {@code NavigationException} is a form of {@link RuntimeException}
+ * that is thrown if errors occur when attempting to determine if
+ * a conditional controller should render or not during navigation.
+ *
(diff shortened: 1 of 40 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/exception/NavigationException.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Constructs a new {@code NavigationException} with a custom message and the specified
+     * cause.
+     *
+     * @param message a custom message
+     * @param cause the cause
+     */
+    public NavigationException(String message, Throwable cause) {
+        super(cause);
```
sroberts2: Should we call `super(message, cause)` so that we don't lose the provided message?

PR #158, opened 2019-01-19
Controller and page for tangible assets note

sroberts2:
Added the Controller for Tangible Assets Note and the webpage.

Helps Resolve: SFA-957

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/TangibleAssetsNoteController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/TangibleAssetsNoteController.java
@@ -1,6 +1,5 @@
 package uk.gov.companieshouse.web.accounts.controller.smallfull;
 
-import java.time.LocalDate;
 import java.util.List;
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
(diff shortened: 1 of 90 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/TangibleAssetsNoteController.java
```diff
@@ -0,0 +1,106 @@
...
+        HttpServletRequest request) {
+
+        addBackPageAttributeToModel(model, companyNumber, transactionId, companyAccountsId);
+
+        if (bindingResult.hasErrors()) {
+            return getTemplateName();
+        }
+
+        try {
+            addDatesToModel(model, companyService.getCompanyProfile(companyNumber));
```
BSaunders95: As discussed, I'd be inclined to add hidden form inputs which take the values of these dates. The dates can be added to the `TangibleAssets` model so that on form submission, the values bind to the model, removing the need for a repeat api call in the case where validation errors occur. 

If binding result errors occur at the moment, these values will be missing from the screen because they haven't been set by that point.

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/TangibleAssetsNoteController.java
```diff
@@ -0,0 +1,105 @@
...
+import uk.gov.companieshouse.web.accounts.model.smallfull.notes.tangible.TangibleAssets;
+import uk.gov.companieshouse.web.accounts.service.company.CompanyService;
+import uk.gov.companieshouse.web.accounts.service.smallfull.TangibleAssetsNoteService;
+import uk.gov.companieshouse.web.accounts.validation.ValidationError;
+
+@Controller
+@NextController(ReviewController.class)
+@PreviousController(OtherAccountingPolicyController.class)
+@RequestMapping("/company/{companyNumber}/transaction/{transactionId}/company-accounts/{companyAccountsId}/small-full/note/tangible-assets")
+public class TangibleAssetsNoteController extends BaseController {
```
caismith-ch: This note should implement the ConditionalController interface as it is an optional note.

PR #159, opened 2019-01-21
Debtors Note Bug Fixes

caismith-ch:
- Update text to match prototype
- Fix for zeros showing on review screen
- Fix conditions for debtors note showing so that it includes checks for zeros
- Fix for previous period being present in mongo when not added

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
@@ -154,9 +154,6 @@ public class BalanceSheetServiceImpl implements BalanceSheetService {
         createCurrentPeriod(apiClient, smallFullApi, currentPeriodUri, currentPeriod,
                 validationErrors);
 
-        checkConditionalNotes(companyProfileApi, balanceSheet, smallFullApi.getLinks(),
-                transactionId, companyAccountsId);
-
         return validationErrors;
 
     }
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/DebtorsServiceImpl.java
```diff
@@ -95,6 +94,23 @@ public Debtors getDebtors(String transactionId, String companyAccountsId, String
         return new ArrayList<>();
     }
 
+    @Override
+    public void deleteDebtors(String transactionId, String companyAccountsId) throws ServiceException {
```
BSaunders95: The delete of the resource should be performed on update of the current / previous resource in the api

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -37,49 +39,61 @@
...
+        BalanceSheet balanceSheet = transformer.getBalanceSheet(currentPeriodApi,
+                previousPeriodApi);
 
         balanceSheet.setBalanceSheetHeadings(balanceSheetHeadings);
 
         return balanceSheet;
     }
 
-    private CurrentPeriodApi getCurrentPeriod(ApiClient apiClient, UriTemplate uri, String transactionId, String companyAccountsId) throws ServiceException {
+    private CurrentPeriodApi getCurrentPeriod(ApiClient apiClient, UriTemplate uri,
```
sroberts2: We don't need to pass `UriTemplate uri` to this private method it is always the `CURRENT_PERIOD_URI` which is available to the whole class.

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -93,10 +107,12 @@ private CurrentPeriodApi getCurrentPeriod(ApiClient apiClient, UriTemplate uri,
         }
     }
 
-    private PreviousPeriodApi getPreviousPeriod(ApiClient apiClient, UriTemplate uri, String transactionId, String companyAccountsId) throws ServiceException {
+    private PreviousPeriodApi getPreviousPeriod(ApiClient apiClient, UriTemplate uri,
```
sroberts2: We don't need to pass `UriTemplate uri` to this private method it is always the `PREVIOUS_PERIOD_URI` which is available to the whole class.

PR #160, opened 2019-01-23
Add delete functionality for debtors to remove note when balancesheet value removed

caismith-ch:
Fix for issue where once a user completes a conditional note and then removes the corresponding values on the balancesheet, the note still exists though the balancesheet value doesn't. Therefore the note will appear on the review screen and be generated in the iXBRL. This PR adds a delete when the note is present and the value in the balancesheet is not.

**Resolves**: SFA-1062

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
@@ -20,6 +20,8 @@ import uk.gov.companieshouse.web.accounts.api.ApiClientService;
 import uk.gov.companieshouse.web.accounts.exception.ServiceException;
 import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheet;
 import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheetHeadings;
+import uk.gov.companieshouse.web.accounts.model.smallfull.CurrentAssets;
+import uk.gov.companieshouse.web.accounts.model.smallfull.Debtors;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -287,4 +290,43 @@ private boolean hasCurrentPeriod(SmallFullLinks smallFullLinks) {
...
+            if ((isDebtorsCurrentAmountNull(balanceSheet) && smallFullLinks.getDebtorsNote() != null)) {
+                debtorsService.deleteDebtors(transactionId, companyAccountsId);
+            }
+        }
+    }
+
+    private boolean isDebtorsCurrentAmountNull(BalanceSheet balanceSheet) {
+        return balanceSheet.getCurrentAssets() != null && balanceSheet.getCurrentAssets().getDebtors() != null
+            && (balanceSheet.getCurrentAssets().getDebtors().getCurrentAmount() == null || balanceSheet.getCurrentAssets().getDebtors().getCurrentAmount() == (0));
+    }
```
sroberts2: It might read better with something like:
`Long currentDebtors = Optional.ofNullable(balanceSheet.getCurrentAssets().getDebtors().getCurrentAmount()).orElse(0L);
return currentDebtors.equals(0L))` 
same with the below method but its personal preference I will leave it up to you.

caismith-ch: Yeah I agree, will update.

caismith-ch: Updated

PR #164, opened 2019-01-28
Add basic web models for stocks - without validation mappings

caismith-ch:
Resolves: SFA-372


ch-mcabrera:
Can we add a description for this PR?

PR #165, opened 2019-01-29
Add POST functionality for creditors after one year web journey

leahtaylor:
Add POST functionality in controller, service and transformer
Add correct navigation for creditors after one year
Add corresponding unit tests

SFA-1053

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/CreditorsAfterOneYearServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/CreditorsAfterOneYearServiceImpl.java
@@ -1,5 +1,7 @@
 package uk.gov.companieshouse.web.accounts.service.smallfull.impl;
 
+import java.util.ArrayList;
+import java.util.List;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
@@ -7,6 +9,7 @@ import org.springframework.web.util.UriTemplate;
 import uk.gov.companieshouse.api.ApiClient;
 import uk.gov.companieshouse.api.error.ApiErrorResponseException;
 import uk.gov.companieshouse.api.handler.exception.URIValidationException;
+import uk.gov.companieshouse.api.model.accounts.smallfull.SmallFullApi;
 import uk.gov.companieshouse.api.model.accounts.smallfull.creditorsafteroneyear.CreditorsAfterOneYearApi;
 import uk.gov.companieshouse.web.accounts.api.ApiClientService;
 import uk.gov.companieshouse.web.accounts.exception.ServiceException;
@@ -15,7 +18,10 @@ import uk.gov.companieshouse.web.accounts.model.smallfull.BalanceSheetHeadings;
 import uk.gov.companieshouse.web.accounts.model.smallfull.notes.creditorsafteroneyear.CreditorsAfterOneYear;
 import uk.gov.companieshouse.web.accounts.service.smallfull.BalanceSheetService;
 import uk.gov.companieshouse.web.accounts.service.smallfull.CreditorsAfterOneYearService;
+import uk.gov.companieshouse.web.accounts.service.smallfull.SmallFullService;
 import uk.gov.companieshouse.web.accounts.transformer.smallfull.CreditorsAfterOneYearTransformer;
+import uk.gov.companieshouse.web.accounts.util.ValidationContext;
+import uk.gov.companieshouse.web.accounts.validation.ValidationError;
 
 @Service
 public class CreditorsAfterOneYearServiceImpl implements CreditorsAfterOneYearService {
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/CreditorsAfterOneYearServiceImpl.java
```diff
@@ -54,6 +66,44 @@ public CreditorsAfterOneYear getCreditorsAfterOneYear(String transactionId,
         return creditorsAfterOneYear;
     }
 
+    @Override
+    public List<ValidationError> submitCreditorsAfterOneYear(String transactionId,
+            String companyAccountsId, CreditorsAfterOneYear creditorsAfterOneYear,
+            String companyNumber) throws ServiceException {
```
BSaunders95: Is `companyNumber` used at all?

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
```diff
@@ -89,8 +109,85 @@ private void populatePreviousPeriodForWeb(CreditorsAfterOneYearApi creditorsAfte
...
     }
 
-    @Override
-    public CreditorsAfterOneYearApi getCreditorsAfterOneYearApi(CreditorsAfterOneYear creditorsAfterOneYear) {
-        return null;
+    private void setCurrentPeriodOnApiModel(CreditorsAfterOneYear creditorsAfterOneYear,
+            CreditorsAfterOneYearApi creditorsAfterOneYearApi) {
+        CurrentPeriod currentPeriod = new CurrentPeriod();
+
+        if (creditorsAfterOneYear.getDetails() != null
```
BSaunders95: You could replace this check with 
```
if (StringUtils.isNotEmpty(creditorsAfterOneYear.getDetails()) {
      currentPeriod.setDetails(creditorsAfterOneYear.getDetails());
}
```

The `isNotEmpty` will catch `null`'s and empty strings (`""`). There's also the option of `isNotBlank` which will also catch whitespace (i.e.: `"       "`).

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/CreditorsAfterOneYearControllerTests.java
```diff
@@ -72,6 +89,8 @@ private void setup() {
 
     private static final String ERROR_VIEW = "error";
 
+    private static final String TEST_PATH = "total.currentTotal";
```
BSaunders95: Can we rename this to be more specific, like `MODEL_VARIABLE` or similar if anyone has a suitable preference?

src/test/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/CreditorsAfterOneYearServiceImplTests.java
```diff
@@ -166,12 +195,233 @@ void getCreditorsAfterOneYearURIValidationException() throws Exception {
...
+
+        when(mockTransformer.getCreditorsAfterOneYearApi(creditorsAfterOneYear)).thenReturn(creditorsAfterOneYearApi);
+
+        creditorsAfterOneYearCreate(creditorsAfterOneYearApi);
+
+        List<ValidationError> validationErrors =
+                creditorsAfterOneYearService.submitCreditorsAfterOneYear(TRANSACTION_ID,
+                COMPANY_ACCOUNTS_ID, creditorsAfterOneYear, COMPANY_NUMBER);
+
+        assertEquals(0, validationErrors.size());
```
BSaunders95: You could `assertTrue(validationErrors.isEmpty())`

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
leahtaylor: So details was storing in Mongo as an empty string - and therefore always showing on the review screen. This method checks if details is not null, but empty - set to null, otherwise, set the details field. Changed to use isBlank rather than "".

```
 if (creditorsAfterOneYear.getDetails() != null
                && StringUtils.isBlank(creditorsAfterOneYear.getDetails())){
            currentPeriod.setDetails(null);
        } else {
            currentPeriod.setDetails(creditorsAfterOneYear.getDetails());
        }
```

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/CreditorsAfterOneYearServiceImpl.java
leahtaylor: removed company number as doesn't seem to be being used

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
BSaunders95: With the logic I suggested above, the value is only ever set on `currentPeriod` if `creditorsAfterOneYear.getDetails()` is not `null` or empty. Since the member variable `details` on `currentPeriod` is `null` by default, this would be sufficient.

At the moment, if `creditorsAfterOneYear.getDetails()` returns `null`, we explicitly call `currentPeriod.setDetails(null)`. It's just not as streamlined as it could be, though I won't block the PR for it

PR #167, opened 2019-01-30
Add creditors after note to review screen

caismith-ch:
Add "Creditors - amounts falling due after one year note" section to the review screen so that a user can see what values they have entered in the note and allow them to return to the note screen to change the values they have provided.

**Resolves**: SFA-1015

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/CreditorsAfterOneYearService.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/CreditorsAfterOneYearService.java
@@ -37,8 +37,7 @@ public interface CreditorsAfterOneYearService {
      *
      * @param transactionId The id of the CHS transaction
      * @param companyAccountsId The company accounts identifier
-     * @return A list of validation errors, or an empty array list if none are present
      * @throws ServiceException if there's an error on deletion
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/CreditorsAfterOneYearService.java
```diff
@@ -30,4 +31,14 @@ CreditorsAfterOneYear getCreditorsAfterOneYear(String transactionId, String comp
...
+
+    /**
+     * Delete the creditors after one year note
+     *
+     * @param transactionId The id of the CHS transaction
+     * @param companyAccountsId The company accounts identifier
+     * @return A list of validation errors, or an empty array list if none are present
+     * @throws ServiceException if there's an error on deletion
+     */
+    List<ValidationError> deleteCreditorsAfterOneYear(String transactionId, String companyAccountsId) throws ServiceException;
```
BSaunders95: This method can be `void`. Delete's should never be 'bad requests'

caismith-ch: 👍

PR #169, opened 2019-01-30
Add tangible assets to the 'check your answers' page

BSaunders95:
- Add tangible assets to the 'check your answers' page
- Move logic to set company dates on the tangible assets object to the service layer

Resolves SFA-544


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/TangibleAssetsNoteServiceImpl.java
```diff
@@ -112,4 +120,14 @@ public TangibleAssets getTangibleAssets(String transactionId, String companyAcco
     private boolean hasTangibleAssetNote(SmallFullLinks smallFullLinks) {
         return smallFullLinks.getTangibleAssetsNote() != null;
     }
+
+    private CompanyProfileApi getCompanyProfile(String companyNumber) throws ServiceException {
+        return companyService.getCompanyProfile(companyNumber);
+    }
+
+    private void addCompanyDatesToTangibleAssets(TangibleAssets tangibleAssets, CompanyProfileApi companyProfile) {
+        tangibleAssets.setLastAccountsPeriodEndOn(companyProfile.getAccounts().getLastAccounts().getPeriodEndOn());
```
JMetherell: Would this fail with a NPE if it was a first year filer?

BSaunders95: It would, yes. We're tackling first year filers in a subsequent story.

PR #170, opened 2019-01-30
Add javascript validation to Tangible note fields

sroberts2:
Added a new fragment `fieldError.html` for individual input errors.
Added this fragment to the `tangibleAssetsNote.html` and the relevant classes for the actual validation.
Updated `baseLayout.html` to remove the balancesheet-validation.js class so that it does not load for pages other than the balancesheet.
Updated the `baseLayout.html` to implement a newer version of the jQuery.

Helps Resolve: SFA-357


caismith-ch:
With the script tag for balancesheet-validation.js being removed from baseLayout.html will break js validation for Debtors as all notes with a balancesheet use that js. Could the script tag be added to the debtors.html file.

PR #176, opened 2019-02-01
Add stocks note to review page

JMetherell:
Add the stocks note to the review screen and remove the note if the corresponding balance sheet fields are emptied or zeroed.

Resolves:
SFA-546


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/StocksServiceImpl.java
```diff
@@ -66,4 +73,29 @@ private StocksApi getStocksApi(String transactionId, String companyAccountsId) t
...
+
+        String uri = STOCKS_URI.expand(transactionId, companyAccountsId).toString();
+
+        try {
+            apiClient.smallFull().stocks().delete(uri).execute();
+        } catch (URIValidationException e) {
+
+            throw new ServiceException(INVALID_URI_MESSAGE, e);
+        } catch (ApiErrorResponseException e) {
+            if (e.getStatusCode() == HttpStatus.BAD_REQUEST.value()) {
```
BSaunders95: It shouldn't be possible to have a bad request when deleting a resource. It should be sufficient to keep this as a void method which can throw a `ServiceException`.

JMetherell: Method is now void

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/BalanceSheetServiceImpl.java
```diff
@@ -308,39 +313,51 @@ private void checkConditionalNotes(CompanyProfileApi companyProfileApi,
...
-            Optional.of(balanceSheet)
+        return Optional.of(balanceSheet)
                 .map(BalanceSheet::getCurrentAssets)
                 .map(CurrentAssets::getDebtors)
                 .map(Debtors::getPreviousAmount)
-                .orElse(0L);
+                .orElse(0L).equals(0L);
+    }
+    
+    private boolean isStocksCurrentAmountNull(BalanceSheet balanceSheet) {
```
BSaunders95: Are these method names a little misleading? A value of 0 would return true, though 0 wouldn't be considered the same as null. Not worth blocking the PR for, just a thought.

BSaunders95:
All looks good. Those method names would be reworked, but the code looks good 👍

PR #186, opened 2019-02-07
Stocks web validation

Mathsh:
Stocks web validation and cross validation

- Updated web models to include @validationMapping 
- Corrected postStocks method in the controller to correct to @ModelAttribute("stocksNote")
- Updated message.properties for stocks
- Added javaScript to stocks

Resolves SFA - 1024 & 876

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/StocksController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/StocksController.java
@@ -60,7 +60,7 @@ public class StocksController extends BaseController implements ConditionalContr
     public String postStocks(@PathVariable String companyNumber,
                               @PathVariable String transactionId,
                               @PathVariable String companyAccountsId,
-                              @ModelAttribute("stocks") @Valid StocksNote stocksNote,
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/StocksController.java
```diff
@@ -60,7 +60,7 @@ public String getStocks(@PathVariable String companyNumber,
     public String postStocks(@PathVariable String companyNumber,
                               @PathVariable String transactionId,
                               @PathVariable String companyAccountsId,
-                              @ModelAttribute("stocks") @Valid StocksNote stocksNote,
+                              @ModelAttribute("stocksNote") @Valid StocksNote stocksNote,
```
ch-mcabrera: we dont need to add the suffix `Note. I would be tempted to remove it as the notes are already in their own package.

Mathsh: it was named after the web model, which is called stocksNote, because it contains stocks.

ch-mcabrera: Ignore me..I've just checked the spec

PR #190, opened 2019-02-11
Adding which account have you prepare controller

ch-mcabrera:
Adding `Which Type Of Accounts Have You Prepared` controller where the user will need to select the account they want to file: 

- Abridged
- Full
- Micros 
- Dormant

Based on the selection the user will be taken to the corresponding page. 

The previous controller has been set to itself until the `corporation tax controller` is completed

Resolve: 
SFA-40


src/main/java/uk/gov/companieshouse/web/accounts/controller/accountselector/SelectAccountTypeController.java
```diff
@@ -0,0 +1,87 @@
...
+        new UriTemplate("/company/{companyNumber}/micro/criteria");
+
+    private static final UriTemplate DORMANT_ACCOUNTS_URI =
+        new UriTemplate("/company/{companyNumber}/dormant/criteria");
+
+    private static final UriTemplate FULL_ACCOUNTS_URI =
+        new UriTemplate("/company/{companyNumber}/small-full/criteria");
+
+    private static final UriTemplate ABRIDGED_ACCOUNTS_URI =
+        new UriTemplate("/company/{companyNumber}");
```
leahtaylor: should this have /abridged-accounts or similar like the other URIs?

PR #193, opened 2019-02-12
Add employees question and note pages

JMetherell:
Add the employees models, service and transformer and the pages for the employees question and employees note

Resolves:
SFA-1087


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/EmployeesController.java
```diff
@@ -0,0 +1,85 @@
...
+      model.addAttribute("employeesNote", employees);
+    } catch (ServiceException e) {
+      LOGGER.errorRequest(request, e.getMessage(), e);
+      return ERROR_VIEW;
+    }
+    return getTemplateName();
+  }
+
+  @PostMapping
+  public String postEmployees(
```
BSaunders95: Do we need to save the employees note as part of this method?

caismith-ch: This will be done in a separate PR.

src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/EmployeesValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    private boolean isEmployeesValid(Employees employees) {
+
+        if (employees == null || (isAverageNumberOfEmployeesNull(employees) &&
+               StringUtils.isEmpty(employees.getDetails()))) {
+            return false;
+        } else {
+            return true;
+        }
```
marcransome: This conditional here can be compressed to a single `return` statement and will silence any associated sonar warnings:

```java
return employees == null || (isAverageNumberOfEmployeesNull(employees) &&
               StringUtils.isEmpty(employees.getDetails());
```

src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/EmployeesValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+    
+    private boolean isAverageNumberOfEmployeesNull(Employees employees) {
+        AverageNumberOfEmployees average = employees.getAverageNumberOfEmployees();
+        if (average == null || (average.getCurrentAverageNumberOfEmployees() == null &&
+             average.getPreviousAverageNumberOfEmployees() == null)) {
+            return true;
+        } else {
+            return false;
+        }
```
marcransome: Similarly, the condition here can be condensed to a single `return` statement:

```java
return average == null || (average.getCurrentAverageNumberOfEmployees() == null &&
               average.getPreviousAverageNumberOfEmployees() == null);
```

PR #199, opened 2019-02-18
Refactor note transformer to accommodate API validation changes

marcransome:
- transform all web model field data API model, rather than conditionally based on the presence of a total, to ensure API field fields are returned and handled
- update note transformers to stop empty periods being included in API requests and avoid unexpected_data errors being returned and blocking the web journey beyond a specific note page
- remove invalid_note in favour of more specific error introduced to API for this purpose (mandatory_element_missing)

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
@@ -109,10 +109,7 @@ public class CreditorsAfterOneYearTransformerImpl implements CreditorsAfterOneYe
             CreditorsAfterOneYearApi creditorsAfterOneYearApi) {
         CurrentPeriod currentPeriod = new CurrentPeriod();
 
-        if (creditorsAfterOneYear.getDetails() != null
-                && StringUtils.isBlank(creditorsAfterOneYear.getDetails())) {
-            currentPeriod.setDetails(null);
-        } else {
+        if (StringUtils.isNotBlank(creditorsAfterOneYear.getDetails())) {
             currentPeriod.setDetails(creditorsAfterOneYear.getDetails());
         }
 
--- a/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
@@ -1,7 +1,5 @@
 package uk.gov.companieshouse.web.accounts.transformer.smallfull.impl;
 
-import java.util.Objects;
-import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
```diff
@@ -115,7 +110,7 @@ private void setCurrentPeriodOnApiModel(CreditorsAfterOneYear creditorsAfterOneY
         CurrentPeriod currentPeriod = new CurrentPeriod();
 
         if (creditorsAfterOneYear.getDetails() != null
-                && StringUtils.isBlank(creditorsAfterOneYear.getDetails())){
+                && StringUtils.isBlank(creditorsAfterOneYear.getDetails())) {
```
BSaunders95: This if / else block can be simplified to:
```
if (StringUtils.isBlank(creditorsAfterOneYear.getDetails())) {
       currentPeriod.setDetails(creditorsAfterOneYear.getDetails());
}
```
Removing the need for the else clause

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/impl/CreditorsAfterOneYearTransformerImpl.java
```diff
@@ -183,12 +180,19 @@ private void setPreviousPeriodOnApiModel(CreditorsAfterOneYear creditorsAfterOne
         }
     }
 
-    private boolean isPreviousPeriodPopulated(PreviousPeriod period) {
-        return Stream.of(
-                period.getBankLoansAndOverdrafts(),
-                period.getFinanceLeasesAndHirePurchaseContracts(),
-                period.getOtherCreditors(),
-           period.getTotal()).anyMatch(Objects ::nonNull);
+    private boolean isCurrentPeriodPopulated(CurrentPeriod currentPeriod) {
```
BSaunders95: As discussed, I think we should perform this check on the `CreditorsAfterOneYear` object to determine whether to create a `CurrentPeriod` in the first place. We can also remove all of the null checks - Thymeleaf handles creation of nested objects for us. But happy to defer those changes to a separate PR.

marcransome: I'll raise this as part of the 'missing tidy up' stories that were raised last week.

src/main/resources/messages.properties
```diff
@@ -155,7 +174,6 @@ validation.length.maxExceeded.accounting_policies.basis_of_measurement_and_prepa
 
 validation.characters.invalid.accounting_policies.turnover_policy = Enter valid characters
 validation.length.minInvalid.accounting_policies.turnover_policy = Give details of your 'turnover policy'
-validation.length.invalidInputLength.accounting_policies.turnover_policy = You must not exceed {1} characters
```
BSaunders95: Did you mean to remove this?

marcransome: No, well spotted. 👍

PR #214, opened 2019-02-23
Removed back links from Balance Sheet and Accounts Chooser

drosser:


src/main/java/uk/gov/companieshouse/web/accounts/controller/accountselector/SelectAccountTypeController.java
```diff
@@ -36,7 +36,6 @@
     public String getTypeOfAccounts(@PathVariable String companyNumber, Model model) {
 
         model.addAttribute("typeOfAccounts", new TypeOfAccounts());
-        addBackPageAttributeToModel(model, companyNumber);
```
BSaunders95: This same change will need to be made on the post method, too. Otherwise, if binding result errors are present on submission, a back link will appear on screen when the page re-renders.

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/BalanceSheetController.java
```diff
@@ -42,8 +42,6 @@ public String getBalanceSheet(@PathVariable String companyNumber,
                                   Model model,
                                   HttpServletRequest request) {
 
-        addBackPageAttributeToModel(model, companyNumber, transactionId, companyAccountsId);
```
BSaunders95: Same here

PR #225, opened 2019-03-05
Add fixed assets investments web validation

JMetherell:
Resolves:
SFA-1124

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/notes/fixedassetsinvestments/FixedAssetsInvestments.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/notes/fixedassetsinvestments/FixedAssetsInvestments.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.web.accounts.model.smallfull.notes.fixedassetsinvestments;
 
+import javax.validation.constraints.NotBlank;
 import lombok.Getter;
 import lombok.Setter;
 import uk.gov.companieshouse.web.accounts.validation.ValidationMapping;
@@ -10,7 +11,8 @@ import uk.gov.companieshouse.web.accounts.validation.ValidationModel;
 @ValidationModel
 public class FixedAssetsInvestments {
 
-    @ValidationMapping("$.fixed_asset_investments.details")
+    @NotBlank(message = "{fixedAssetsInvestments.details.missing}")
+    @ValidationMapping("$.fixed_assets_investments.details")
     private String details;
 
 }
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/model/smallfull/notes/fixedassetsinvestments/FixedAssetsInvestments.java
```diff
@@ -10,7 +11,8 @@
 @ValidationModel
 public class FixedAssetsInvestments {
 
-    @ValidationMapping("$.fixed_asset_investments.details")
+    @NotBlank(message = "{fixedAssetsInvestments.details.missing}")
```
BSaunders95: This isn't required since the api validation enforces this. This can be removed, along with the corresponding ValidationMessage property.

src/main/resources/messages.properties
```diff
@@ -130,6 +130,9 @@ validation.element.missing.creditors_after_one_year.previous_period.total = Ente
 validation.unexpected.data.creditors_after_one_year.current_period = The total figure here must match the 'Creditors - amounts falling due after one year' figure on the balance sheet
 validation.unexpected.data.creditors_after_one_year.previous_period = The total figure here must match the 'Creditors - amounts falling due after one year' figure on the balance sheet
 
+#Fixed assets investments
+validation.characters.invalid.fixed_assets_investments.details = Invalid characters entered
+validation.length.invalidInputLength.fixed_assets_investments.details = Field must not exceed 20,000 characters
```
BSaunders95: As a result of the api validation, this message should read `Field must not exceed {1} characters`. The validation mapping utility in the web will then read index 1 of the `Error` object's arguments, which will be 20000. Better to have a placeholder here in case the upper limit on the length of the string ever changes.

JMetherell: Updated to use the placeholder

src/main/resources/ValidationMessages.properties
```diff
@@ -7,3 +7,4 @@ otherAccountingPolicy.selectionNotMade=Tell us if your prepared accounts include
 valuationInformationPolicy.selectionNotMade=Tell us if your prepared accounts include a ''valuation information'' policy
 corporationTax.selectionNotMade=Tell us if you also want to file your corporation tax with HMRC now
 typeOfAccounts.selectionNotMade=Tell us which type of accounts you have prepared
+fixedAssetsInvestments.details.missing = Provide details of your 'fixed assets investments'
```
BSaunders95: Will this message need double single-quotes like the others?

JMetherell: Yes, it should have.  Double single quotes added.

PR #232, opened 2019-03-12
Update numberInput fragment to display date in mobile labels

leahtaylor:
Update mobile labels on number input fields on balance sheet and notes to display correctly as

'{field} {period} {period date} in £'

SFA-1287


src/main/resources/templates/fragments/numberInput.html
```diff
@@ -1,20 +1,19 @@
...
-                     type="text">
-            </label>
-        </div>
-    </body>
+<head>
+  <title></title>
+</head>
+<body>
+<div th:fragment="numberInput" class="govuk-form-group">
+  <label th:for="${id}">
```
Mathsh: was this meant to be removed?

 <span class="currency-type">£</span>

PR #239, opened 2019-03-18
Fixed errors in Employees Note

Mathsh:
Corrected employees note

- Removed duplicate ids by including employees in the id name.
- Updated Form class to remove currency-included tag as not a currency
- Added th:if"${isEmployees !=true} to numberInput to remove £ when employees note
- Updated employees question html to include 'employees'

Resolves SFA 1108 & 1087

### DIFF
--- a/src/main/resources/templates/smallfull/employees.html
+++ b/src/main/resources/templates/smallfull/employees.html
@@ -55,7 +55,7 @@
         <div class="govuk-grid-column-one-half">
 
           <details class="govuk-details" role="group">
-            <summary class="govuk-details__summary" role="button" aria-controls="details-content"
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/smallfull/employees.html
```diff
@@ -57,13 +57,13 @@ <h2 class="govuk-heading-s">
           <details class="govuk-details" role="group">
             <summary class="govuk-details__summary" role="button" aria-controls="details-content"
```
JMetherell: Does this need to be chnaged to employees-details-content to match?

Mathsh: Updated to match

PR #240, opened 2019-03-18
Add gov.uk journey for smallfull accounts

caismith-ch:
Adds new gov.uk journey. Additonal controllers where added for use of existing templates as the current build of navigation requires class level request mappings. This prevents multiple requestmappings as different functionality for POST methods is required. Class level requestmappings prevents differentiation between methods and the request mapping.

**Resolves**: SFA-1195, SFA-1251, SFA-38


src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukSelectAccountTypeController.java
```diff
@@ -0,0 +1,82 @@
...
+        if (bindingResult.hasErrors()) {
+            return getTemplateName();
+        }
+
+        return getReDirectPageURL(typeOfAccounts.getSelectedAccountTypeName());
+    }
+
+    private String getReDirectPageURL(String selectedAccount) {
+
+        if ("micros".equalsIgnoreCase(selectedAccount)) {
```
BSaunders95: We could think about using a `StringBuilder` and a switch case statement to build the returned String, rather than multiple 'if's, but that's a matter of preference

src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukSelectAccountTypeController.java
```diff
@@ -0,0 +1,82 @@
...
+
+        if ("full".equalsIgnoreCase(selectedAccount)) {
+            return UrlBasedViewResolver.REDIRECT_URL_PREFIX + FULL_ACCOUNTS_URI.toString();
+        }
+
+        if ("dormant".equalsIgnoreCase(selectedAccount)) {
+            return UrlBasedViewResolver.REDIRECT_URL_PREFIX + dormantAccountsUri;
+        }
+
+        return getTemplateName();
```
BSaunders95: Does this make sense? Should we re-render the page if a user ever makes it this far? Granted, this should never be possible, but we could cater for the error by throwing a RuntimeException. That way, we'll handle the error because the `GlobalExceptionHandler` will catch the error, log relevant details and render the error page.

src/main/java/uk/gov/companieshouse/web/accounts/controller/fullaccounts/FullAccountsCriteriaController.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @GetMapping
+    public String getCriteria(Model model){
+        model.addAttribute("hideUserBar" , true);
+        return getTemplateName();
+    }
+
+    @PostMapping
+    public String postFullAccountsCriteria(){
+        return UrlBasedViewResolver.REDIRECT_URL_PREFIX + chsUrl;
```
BSaunders95: Does this journey redirect a user to the CHS search page?

caismith-ch: Yes it does following the prototype. In the future it will likely point to the Company Number page which has yet to be built.

src/main/java/uk/gov/companieshouse/web/accounts/controller/fullaccounts/FullAccountsCriteriaController.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    protected String getTemplateName() {
+        return "fullaccounts/criteria";
```
BSaunders95: Maybe we could nest the templates slightly differently. With these changes we'd have both a 'fullaccounts' and a 'smallfull' directory listed under templates. I'd suggest maybe having a structure like 
```
templates
  |
  |__ smallfull
  |
  |__ govuk
        |
        |__ smallfull
```
Just to maintain consistency?

src/main/java/uk/gov/companieshouse/web/accounts/controller/fullaccounts/FullAccountsCriteriaController.java
BSaunders95: Fair enough!

src/main/java/uk/gov/companieshouse/web/accounts/controller/fullaccounts/FullAccountsCriteriaController.java
caismith-ch: Agreed that makes more sense.

src/main/java/uk/gov/companieshouse/web/accounts/controller/fullaccounts/FullAccountsCriteriaController.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.servlet.view.UrlBasedViewResolver;
+import uk.gov.companieshouse.web.accounts.controller.BaseController;
+
+@Controller
+@RequestMapping("/accounts/full-accounts-criteria")
+public class FullAccountsCriteriaController extends BaseController{
```
BSaunders95: Maybe worth renaming this to `GovukFullAccountsCriteriaController` for consistency with the other controllers in this PR?

caismith-ch: 👍

src/main/java/uk/gov/companieshouse/web/accounts/CompanyAccountsWebApplication.java
```diff
@@ -36,7 +36,7 @@ public static void main(String[] args) {
     public void addInterceptors(InterceptorRegistry registry) {
 
         registry.addInterceptor(loggingInterceptor);
-        registry.addInterceptor(userDetailsInterceptor);
-        registry.addInterceptor(companyAccountsDataStateInterceptor);
+        registry.addInterceptor(userDetailsInterceptor).excludePathPatterns("/accounts/*");
+        registry.addInterceptor(companyAccountsDataStateInterceptor).excludePathPatterns("/accounts/*");;
```
BSaunders95: There's a trailing ';' which can be removed

src/test/java/uk/gov/companieshouse/web/accounts/controller/fullaccounts/FullAccountsCriteriaControllerTest.java
```diff
@@ -0,0 +1,64 @@
...
+            .andExpect(model().attributeExists(TEMPLATE_NAME_MODEL_ATTR));
+    }
+
+    @Test
+    @DisplayName("Post corporation tax with criteria met")
+    void postRequestCriteriaMet() throws Exception {
+
+        ReflectionTestUtils.setField(controller, "chsUrl", "/");
+        this.mockMvc.perform(post(PATH)
+            .param("chsUrl" ,"/"))
```
BSaunders95: Is this param required?

src/test/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukSelectAccountTypeControllerTest.java
```diff
@@ -0,0 +1,140 @@
...
+    private MockMvc mockMvc;
+
+    @Mock
+    private NavigatorService mockNavigatorService;
+
+    @InjectMocks
+    private GovukSelectAccountTypeController controller;
+
+    @BeforeEach
+    void setUpBeforeEAch() {
```
BSaunders95: Typo on the capital 'A'

src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukSelectAccountTypeController.java
caismith-ch: Gave it a go, sonarlint complains about readability using a switch which I kind of agree with slightly. Will keep the same for this instance.

BSaunders95: That's cool, it really was a styling choice and it performs its function the way it is 👍

caismith-ch: Appreciate the suggestion 👍

src/main/resources/templates/fragments/userBar.html
```diff
@@ -4,7 +4,7 @@
         <title>User Bar</title>
     </head>
     <body>
-        <div th:fragment="userBar" class="js-toggle-nav" id="global-nav">
+        <div th:fragment="userBar" th:classappend="${hideUserBar != null ? 'hidden' : ''}" class="js-toggle-nav" id="global-nav">
```
BSaunders95: Just a minor thought, rather than passing a new attribute to the fragment, could we conditionally display the user bar based on whether `userEmail` is null?

src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukCorporationTaxController.java
```diff
@@ -22,8 +24,11 @@
     @GetMapping
     public String getCorporationTax(Model model) {
 
+        EnvironmentReader enviromentReader = new EnvironmentReaderImpl();
+
         model.addAttribute("corporationTax", new CorporationTax());
         model.addAttribute("hideUserBar", true);
+        model.addAttribute("backButton", enviromentReader.getMandatoryString("GOV_UK_FULL_ACCOUNTS_URL"));
```
BSaunders95: Could we make `"GOV_UK_FULL_ACCOUNTS_URL"` a constant rather than using a sting literal here?

caismith-ch: This has changed now as we are getting the value from the application.properties file to be consistent with the other external urls used in the application.

src/main/resources/application.properties
```diff
@@ -10,3 +10,4 @@ piwik.siteId=${PIWIK_SITE_ID}
 micro-entity-accounts.uri=https://beta.companieshouse.gov.uk/guides/accounts/chooser/micro-entity
 dormant-accounts.uri=https://beta.companieshouse.gov.uk/guides/accounts/chooser/dormant
 abridged-accounts.uri=https://beta.companieshouse.gov.uk/guides/accounts/chooser/abridged
+gov-uk-full-accounts.uri=http://www.gov.uk/file-your-company-annual-accounts
```
BSaunders95: Is this url accounts specific?

src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukCorporationTaxController.java
```diff
@@ -20,14 +20,13 @@
 @RequestMapping("/accounts/corporation-tax")
 public class GovukCorporationTaxController extends BaseController {
 
-    @Value("${gov-uk-full-accounts.uri}")
+    @Value("${gov-uk-file-your-accounts.uri}")
```
BSaunders95: Since this has changed it may also be worth changing the name of the variable too?

caismith-ch: 💯

PR #243, opened 2019-03-19
include called up share capital field in balance sheet calculations

leahtaylor:
add extra class to include called up share capital field in javascript total calculation for total net assets less current liabilities field

SFA-1308


CarlSwan:
CHS Release - REL0001459 / CHG0046660

PR #259, opened 2019-04-05
Change body of company accounts POST request to be empty in line with API changes

BSaunders95:
Change body of company accounts POST request to be empty in line with API changes; accounting period dates are now derived from the company profile associated with the transaction used to create the company accounts.

Resolves: SFA-1360


marcransome:
Can you update the `pom.xml` to reference the new `api-sdk-java` build version `3.13.27`?

PR #283, opened 2019-05-15
Add initial CIC steps to complete code

drosser:
Previous and Next Controller both go to SFA Steps to Complete screen until the relevant screens are also built


src/main/java/uk/gov/companieshouse/web/accounts/service/transaction/impl/TransactionServiceImpl.java
```diff
@@ -12,6 +12,8 @@
 import uk.gov.companieshouse.web.accounts.exception.ServiceException;
 import uk.gov.companieshouse.web.accounts.service.transaction.TransactionService;
 
+import java.util.Optional;
```
BSaunders95: Unused import

drosser: Updated.

BSaunders95: 👍

PR #286, opened 2019-05-21
Update Web to Use SDK V4

BSaunders95:
Update the application to use version 4.0.0 of the API SDK following a breaking change.

Resolves SFA-1695

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/companyaccounts/impl/CompanyAccountsServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/companyaccounts/impl/CompanyAccountsServiceImpl.java
@@ -17,7 +17,7 @@ import uk.gov.companieshouse.web.accounts.service.companyaccounts.CompanyAccount
 public class CompanyAccountsServiceImpl implements CompanyAccountsService {
 
     @Autowired
-    ApiClientService apiClientService;
+    private ApiClientService apiClientService;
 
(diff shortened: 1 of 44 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/companyaccounts/impl/CompanyAccountsServiceImpl.java
```diff
@@ -34,7 +34,7 @@ public String createCompanyAccounts(String transactionId) throws ServiceExceptio
         String uri = CREATE_COMPANY_ACCOUNTS_URI.expand(transactionId).toString();
 
         try {
-            companyAccounts = apiClient.companyAccounts().create(uri, companyAccounts).execute();
+            companyAccounts = apiClient.companyAccounts().create(uri, companyAccounts).execute().getData();
```
caismith-ch: Do we need to getData() on a create?

BSaunders95: We do in this case because we pull back the response body and query the self link to fetch the `companyAccountsId`

caismith-ch: Cool, just wanted to check as it was different to the others 👍

PR #297, opened 2019-05-24
CIC Approval page

sroberts2:
Added Directors approval page for the CIC report.

Resolves: SFA-1634


src/main/java/uk/gov/companieshouse/web/accounts/controller/cic/CicApprovalController.java
```diff
@@ -0,0 +1,99 @@
...
+    @GetMapping
+    public String getApproval(@PathVariable String companyNumber,
+        @PathVariable String transactionId,
+        @PathVariable String companyAccountsId,
+        Model model) {
+
+        addBackPageAttributeToModel(model, companyNumber, transactionId, companyAccountsId);
+
+        model.addAttribute(APPROVAL, new CicApproval());
+        model.addAttribute(TRANSACTION_ID, transactionId);
```
BSaunders95: These attributes won't be required

src/main/java/uk/gov/companieshouse/web/accounts/controller/cic/CicApprovalController.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @GetMapping
+    public String getApproval(@PathVariable String companyNumber,
+        @PathVariable String transactionId,
+        @PathVariable String companyAccountsId,
+        Model model) {
+
+        addBackPageAttributeToModel(model, companyNumber, transactionId, companyAccountsId);
+
+        model.addAttribute(APPROVAL, new CicApproval());
```
BSaunders95: We should try to fetch an existing approval here because a user may have already provided one, rather than providing a new instance each time.

src/main/java/uk/gov/companieshouse/web/accounts/service/cic/impl/CicApprovalServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+            return serviceExceptionHandler.handleSubmissionException(e, RESOURCE_NAME);
+        } catch (URIValidationException e) {
+            serviceExceptionHandler.handleURIValidationException(e, RESOURCE_NAME);
+        }
+
+        return new ArrayList<>();
+    }
+
+    @Override
+    public List<ValidationError> validateCicApprovalDate(CicApproval cicApproval) {
```
BSaunders95: All of this logic already exists in the `ApprovalServiceImpl` class, we should pull this into a generic validator class

src/main/resources/templates/cic/cicApproval.html
```diff
@@ -0,0 +1,147 @@
...
+                           id="approval-date-year"
+                           th:field="*{date.year}"
+                           type="text">
+                  </div>
+                </div>
+
+              </div>
+            </fieldset>
+
+            <div class="govuk-inset-text onlyJS">
```
BSaunders95: We won't display the widget on this page

PR #304, opened 2019-06-04
Add branching navigation

drosser:


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/StepsToCompleteController.java
```diff
@@ -42,30 +44,54 @@ protected String getTemplateName() {
...
+            } catch (ServiceException e) {
 
-            statementsService.createBalanceSheetStatementsResource(transactionId, companyAccountsId);
+                LOGGER.errorRequest(request, e.getMessage(), e);
+                return ERROR_VIEW;
+            }
+        }
 
-            transactionService.createResumeLink(companyNumber, transactionId, companyAccountsId);
+        try {
```
BSaunders95: You can wrap the whole thing in a single try / catch; the catch behaves the same in both blocks

src/main/java/uk/gov/companieshouse/web/accounts/service/navigation/NavigatorService.java
```diff
@@ -77,13 +116,51 @@ private Class getNextControllerClass(Class clazz) {
...
+    private Class getPreviousControllerClass(Class clazz, String ... pathVars) {
         Annotation previousControllerAnnotation = AnnotationUtils.findAnnotation(clazz, PreviousController.class);
         if (previousControllerAnnotation == null) {
             throw new MissingAnnotationException("Missing @PreviousController annotation on " + clazz.toString());
         }
 
-        return ((PreviousController) previousControllerAnnotation).value();
+        Class[] classList = ((PreviousController) previousControllerAnnotation).value();
+
+        if (classList.length > 0) {
```
BSaunders95: This logic is the same as the above. Could it be moved into a common method?

drosser: This is a pain due to the casting of some objects that takes place

PR #305, opened 2019-06-05
Added Govuk company details

amoriartyCH:
- Added new controller to expose Gov UK company details page
- Updated company profile service to pull `CompanyDetails` from company profile
- Added new transformer

This change introduced a company details page which would previously belong to the lookup service. The reason for moving it is to allow flexibility in the display on this data.

[SFA - 1578]

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukCompanyDetailController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukCompanyDetailController.java
@@ -1,0 +1,52 @@
+package uk.gov.companieshouse.web.accounts.controller.govuk;
+
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/govuk/GovukCompanyDetailController.java
```diff
@@ -0,0 +1,52 @@
...
+    protected String getTemplateName() {
+        return "company/companyDetail";
+    }
+
+    @GetMapping
+    public String getCompanyDetail(@PathVariable String companyNumber, Model model, HttpServletRequest request) {
+
+        try {
+            model.addAttribute("companyDetail", companyService.getCompanyDetail(companyNumber));
+            model.addAttribute("templateHeading", TEMPLATE_HEADING);
```
drosser: Pedantic: This doesn't need to be inside the `try/catch`

PR #333, opened 2019-08-30
Feature/intangible assets controller

Jamie87:
Added intangible assets controller

### DIFF
--- a/src/main/resources/templates/smallfull/intangibleAssetsNote.html
+++ b/src/main/resources/templates/smallfull/intangibleAssetsNote.html
@@ -1,0 +1,354 @@
+<!DOCTYPE html>
+<html lang="en" xmlns:th="http://www.thymeleaf.org"
+      xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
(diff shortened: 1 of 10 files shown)


src/main/resources/templates/smallfull/intangibleAssetsNote.html
```diff
@@ -0,0 +1,303 @@
...
+    <div th:replace="fragments/numberedHeading :: numberedHeading (
+                    headingText = 'Intangible assets note'
+                )">
+    </div>
+
+    <input th:field="*{lastAccountsPeriodEndOn}" type="hidden">
+    <input th:field="*{nextAccountsPeriodStartOn}" type="hidden">
+    <input th:field="*{nextAccountsPeriodEndOn}" type="hidden">
+
+    <p class="warning-note warning-note-large">You only need to enter items included in your
```
BSaunders95: This <p> tag isn't required to match the prototype

src/main/resources/templates/smallfull/intangibleAssetsNote.html
```diff
@@ -0,0 +1,303 @@
...
+      <br>
+
+    </div>
+
+
+      <br>
+
+    </th:block>
+
+    <div class="form-group">
```
BSaunders95: I don't think this wrapping div is needed

PR #338, opened 2019-09-05
Add javascript classes to html

Jamie87:
Added javascript classes to intangibleAssetsNote.html to allow for totalling


BSaunders95:
From what I understand, these changes will only total across rows at present

PR #349, opened 2019-09-20
Feature/intangible nbv javascript totalling

faisal-amin82:
-Added javascript totalling classes for net book values


faisal-amin82:
Opened in error

PR #360, opened 2019-10-11
add div classes to balancesheet page

Jamie87:
Added div classes to balanceSheet.html to fix alignment issues

### DIFF
--- a/src/main/resources/templates/smallfull/balanceSheet.html
+++ b/src/main/resources/templates/smallfull/balanceSheet.html
@@ -27,18 +27,18 @@
                     <!-- Hidden input for form binding -->
                     <input th:field="*{lbg}" type="hidden">
 
-                    <div class="govuk-grid-row" id="accounts-header">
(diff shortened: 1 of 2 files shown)


src/main/resources/templates/smallfull/balanceSheet.html
```diff
@@ -27,18 +27,18 @@
...
-                        <div class="govuk-grid-column-one-quarter">
+                        <div class="govuk-grid-column-one-quarter column-flex">
                             <h2 class="govuk-heading-s">
                                 <span id="currentPeriodHeading" th:text='*{balanceSheetHeadings.currentPeriodHeading}'></span>
                                 <!-- Hidden input for form binding -->
                                 <input th:field="*{balanceSheetHeadings.currentPeriodHeading}" type="hidden">
                             </h2>
                         </div>
-                        <div class="govuk-grid-column-one-quarter">
+                        <div class="govuk-grid-column-one-quarter column-flex">
```
BSaunders95: Probably don't need to worry about the headings above

PR #374, opened 2019-10-31
added validation model annotation

faisal-amin82:
added ValidationModel annotation to the ProfitAndLoss class for correct validation mapping


Jamie87:
Changes already in previous base branch

PR #388, opened 2019-11-12
Disable save and continue after first click to avoid multiple clicks

faisal-amin82:
- added code to disable multiple clicks on a save and continue button so that once user has submitted the resource, it won't resubmit the same resource again

Resolves: SFA-2298


faisal-amin82:
Not required

PR #390, opened 2019-11-15
Add P&L to CYA Page

BSaunders95:
Resolves SFA-2122

### DIFF
--- a/src/main/resources/templates/smallfull/review.html
+++ b/src/main/resources/templates/smallfull/review.html
@@ -19,14 +19,64 @@
 
     <!-- Start Profit and Loss-->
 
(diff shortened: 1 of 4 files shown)


src/main/resources/templates/smallfull/review.html
```diff
@@ -16,6 +16,157 @@ <h1 id="page-title-heading" class="govuk-heading-l">Check your answers</h1>
...
+                        )">
+          </div>
+
+        </th:block>
+
+      </dl>
+
+    </div>
+
+    <!-- End Profit and Loss
```
Jamie87: Have you closed off the comment properly here?

BSaunders95: Good catch, addressed

PR #415, opened 2020-01-15
statement page fix

Jamie87:
Fixes persisting bullet point on Balance sheet statements page.
Resolves SFA-2042


src/main/resources/templates/smallfull/statements.html
```diff
@@ -32,7 +32,9 @@ <h1 id="page-title" class="govuk-heading-l">
                         <li id="statements-audit-not-required-by-members" th:text="*{auditNotRequiredByMembers}"></li>
                         <li id="statements-directors-responsibility" th:text="*{directorsResponsibility}"></li>
                         <li id="statements-small-companies-regime" th:text="*{smallCompaniesRegime}"></li>
+                        <div th:if="*{noProfitAndLoss} != null">
```
BSaunders95: You can actually apply the `th:if` to the `li` element to save the redundant div. Also, the '!= null' can live within the curly braces.
Alternatively, if you opt not to apply the if to the `li` tag, replace the div with `th:block` - a meta-html element which doesn't change the unlying rendered html.

PR #431, opened 2020-02-18
edit makefile and add version file

Jamie87:
Resolves 
SFA2589


version
```diff
@@ -0,0 +1 @@
+1.2
```
BSaunders95: Should this be 1.13?

Jamie87: Totally right

PR #436, opened 2020-02-25
Remove employees question

Jamie87:
Ensures the employees page is mandatory by removing the question page.

Resolves:
SFA- 2265


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/EmployeesController.java
```diff
@@ -26,7 +26,7 @@
 
 @Controller
 @NextController(IntangibleAssetsNoteController.class)
-@PreviousController(EmployeesQuestionController.class)
+@PreviousController(OtherAccountingPolicyController.class)
```
BSaunders95: The `EmployeesController` class should no longer implement `ConditionalController`; we will always render the page going forward.

BSaunders95: The `EmployeesController` class should no longer implement `ConditionalController`; we will always render the page going forward.

BSaunders95:
Be sure to remove the `EmployeesQuestion` class, as well as any `ValidationMessages` relevant to the employees question

PR #438, opened 2020-02-26
Add mandatory validation for employees web

faisal-amin82:
- Add mandatory validation for average number of employees when the values are missing

Resolves: SFA-2590


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/EmployeesController.java
```diff
@@ -19,7 +19,6 @@
 import uk.gov.companieshouse.web.accounts.model.state.CompanyAccountsDataState;
 import uk.gov.companieshouse.web.accounts.service.smallfull.EmployeesService;
 import uk.gov.companieshouse.web.accounts.validation.ValidationError;
-import uk.gov.companieshouse.web.accounts.validation.smallfull.EmployeesValidator;
```
BSaunders95: Do you want to delete the actual class at the same time as we do this work?

PR #478, opened 2020-06-02
Hide future dates from user logs in after periodEndOn

mreis-ch:
Hide future dates when user logs into service after period end date up to 7 days. Update method getFutureDatesForArd to loop through 7 days from periodEndOn and add to list if the new date is before or on today's date.

BI-3186

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/company/impl/CompanyServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/company/impl/CompanyServiceImpl.java
@@ -128,16 +128,15 @@ public class CompanyServiceImpl implements CompanyService {
     @Override
     public List<LocalDate> getFutureDatesForArd(LocalDate periodEndOn) {
 
+    	LocalDate todaysDate = LocalDate.now();
         List<LocalDate> futureValidDates = new ArrayList<>();
 
-
-        futureValidDates.add(periodEndOn.plusDays(1));
-        futureValidDates.add(periodEndOn.plusDays(2));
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/company/impl/CompanyServiceImpl.java
```diff
@@ -128,16 +128,15 @@ private boolean isSameYearFiler(CompanyProfileApi companyProfile) {
...
-
-        futureValidDates.add(periodEndOn.plusDays(1));
-        futureValidDates.add(periodEndOn.plusDays(2));
-        futureValidDates.add(periodEndOn.plusDays(3));
-        futureValidDates.add(periodEndOn.plusDays(4));
-        futureValidDates.add(periodEndOn.plusDays(5));
-        futureValidDates.add(periodEndOn.plusDays(6));
-        futureValidDates.add(periodEndOn.plusDays(7));
+        for(int i = 1; i <= 7; i++) {
+        	if(periodEndOn.plusDays(i).isBefore(todaysDate) ||
```
BSaunders95, line 135: How about flipping this logic:
```
if (periodEndOn.plusDays(i).isAfter(todaysDate)) {
  break; // break to bail out of the loop early to improve performance
}
futureValidDates.add(periodEndOn.plusDays(i));
```

mreis-ch: Sounds a reasonable switch around. Will update

PR #485, opened 2020-06-04
Refactor current assets investments to use new NoteService pattern

BSaunders95:
Resolves BI-3561


src/main/java/uk/gov/companieshouse/web/accounts/service/notehandler/smallfull/CurrentAssetsInvestmentsHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import uk.gov.companieshouse.web.accounts.service.smallfull.SmallFullService;
+
+@Component
+public class CurrentAssetsInvestmentsHandler implements
+        NoteResourceHandler<CurrentAssetsInvestmentsApi> {
+
+    @Autowired
+    private SmallFullService smallFullService;
+
+    private static final UriTemplate STOCKS_URI =
```
mreis-ch, line 23: Should STOCKS_URI be called something like CURRENT_ASSETS_INVESTMENTS_URI for consistency?

BSaunders95: Good catch! Yes it should - renamed

PR #493, opened 2020-06-09
Intangible Assets: use new NoteService,NoteType

ch-tstrickland:
To save unnecessary lines of code, the NoteService is a new area for
 common code, previously in classes such as
 IntangibleAssetsNoteService(Impl). This commit uses the new NoteService
 and deletes the old code that the NoteService replaces.
Add Intangible Assets type to NoteType.
Create IntangibleAssetHandler, IntangibleAssetDateHandler classes.
Update tests, add new tests.

Fixes BI-3565


src/main/java/uk/gov/companieshouse/web/accounts/service/notehandler/smallfull/IntangibleAssetsHandler.java
```diff
@@ -0,0 +1,62 @@
...
+import uk.gov.companieshouse.web.accounts.service.notehandler.NoteResourceHandler;
+import uk.gov.companieshouse.web.accounts.service.smallfull.SmallFullService;
+
+@Component
+public class IntangibleAssetsHandler implements NoteResourceHandler<IntangibleApi> {
+
+    @Autowired
+    private SmallFullService smallFullService;
+
+    private static final UriTemplate EMPLOYEES_URI = new UriTemplate(
```
BSaunders95, line 22: Should this be `INTANGIBLE_URI`?

src/main/java/uk/gov/companieshouse/web/accounts/service/notehandler/smallfull/IntangibleAssetsHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        return EMPLOYEES_URI.expand(transactionId, companyAccountsId).toString();
+    }
+
+    @Override
+    public Executor<ApiResponse<IntangibleApi>> get(ApiClient apiClient, String uri) {
+        return apiClient.smallFull().intangible().get(uri);
+    }
+
+    @Override
+    public Executor<ApiResponse<Void>> update(ApiClient apiClient, String uri, IntangibleApi employeesApi) {
```
BSaunders95, line 37: Should the api object be renamed?

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/intangible/impl/IntangibleAssetsTransformerImpl.java
```diff
@@ -7,19 +7,20 @@
...
 @Component
-public class IntangibleAssetsTransformerImpl implements IntangibleAssetsTransformer {
+public class IntangibleAssetsTransformerImpl implements NoteTransformer<IntangibleAssets, IntangibleApi> {
 
     @Autowired
     private IntangibleAssetsTransformerFactory factory;
 
     @Override
-    public IntangibleAssets getIntangibleAssets(IntangibleApi intangibleApi) {
+    public IntangibleAssets toWeb(IntangibleApi intangibleApi) {
```
BSaunders95, line 23: We need an additional check in this method for the case where `intangibleApi == null`

src/main/java/uk/gov/companieshouse/web/accounts/service/notehandler/smallfull/IntangibleAssetsHandler.java
amoriartyCH, line 22: should this be intangible_uri

amoriartyCH: Just saw Brad has commented, will resolve this comment as its a duplicate.

src/main/java/uk/gov/companieshouse/web/accounts/transformer/smallfull/intangible/impl/IntangibleAssetsTransformerImpl.java
ch-tstrickland: > We need an additional check in this method for the case where `intangibleApi == null`

... which would then return null? Or just empty?

BSaunders95: If the api object is null, we need to return an 'empty' object to back the page - see other transformers for examples 👍

ch-tstrickland: Fixed.

PR #497, opened 2020-06-11
Accounting Policies rework for web

amoriartyCH:
-Updated the Accounting Policies note to use the shared note service.
-Changed code to fit within new structure of the re-work thats currently in-progress.


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/ReviewServiceImpl.java
```diff
@@ -116,20 +96,23 @@ public Review getReview(String transactionId, String companyAccountsId, String c
 
         Statements statements = statementsService.getBalanceSheetStatements(transactionId, companyAccountsId);
 
-        BasisOfPreparation basisOfPreparation = basisOfPreparationService.getBasisOfPreparation(transactionId, companyAccountsId);
+        BasisOfPreparation basisOfPreparation = accountingPoliciesNoteService.get(transactionId, companyAccountsId, NoteType.SMALL_FULL_ACCOUNTING_POLICIES)
```
BSaunders95, line 99: How about we call `accountingPoliciesNoteService.get(transactionId, companyAccountsId, NoteType.SMALL_FULL_ACCOUNTING_POLICIES)` once and assign that to an `AccountingPolicies` variable, then call 'gets' on that one object to cut 6 API calls down to 1?

amoriartyCH: Sounds great, pushed changes. Resolving this issue.

PR #503, opened 2020-06-22
Resolve reappointed directors approval issue

faisal-amin82:
- Resolves issue where directors with appointment date after the resignation date are treated as resigned and filtered out from approval page. This fix treats those directors as reappointed and not resigned so that those are not filtered out

Resolves: SFA-3217

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/DirectorsReportApprovalController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/DirectorsReportApprovalController.java
@@ -69,8 +69,19 @@ public class DirectorsReportApprovalController extends BaseController implements
                     directorsReportApprovalService.getDirectorsReportApproval(transactionId, companyAccountsId);
 
             List<String> approverOptions =
-                    Arrays.stream(directorService.getAllDirectors(transactionId, companyAccountsId))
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/DirectorsReportApprovalController.java
```diff
@@ -69,8 +69,19 @@ public String getDirectorsReportApproval(@PathVariable String companyNumber,
                     directorsReportApprovalService.getDirectorsReportApproval(transactionId, companyAccountsId);
 
             List<String> approverOptions =
-                    Arrays.stream(directorService.getAllDirectors(transactionId, companyAccountsId))
-                            .filter(d -> d.getResignationDate() == null)
+                    Arrays.stream(directorService.getAllDirectors(transactionId, companyAccountsId)).map(
```
ch-tstrickland, line 72: I have only one Q here, but it may be a bit larger than this PR. What is the largest size of reply for getAllDirectors? Would it (should it) ever be paged?

BSaunders95: Max is 20 so we never worry about paging!

PR #507, opened 2020-07-23
Update basis of preparation page

faisal-amin82:
- Update basis of preparation page to include FRS101 as a regulatory standard along with FRS102

Resolves: BI- 3277


src/main/resources/templates/smallfull/basisOfPreparation.html
```diff
@@ -44,18 +44,19 @@ <h1 id="page-title-heading" class="govuk-heading-l govuk-!-margin-bottom-0">How
                             </span>
 
               <div class="govuk-radios__item">
-                <input class="govuk-radios__input piwik-event" id="basisOfPreparation-prepared-statement" data-target-text-field="basisOfPreparation-custom-statement-field" th:field="*{isPreparedInAccordanceWithStandards}" type="radio" value="1" data-event-id="FRS102">
-                <label class="govuk-label govuk-radios__label" id="basisOfPreparation-prepared-statement-label" for="basisOfPreparation-prepared-statement">
-                  These financial statements have been prepared in accordance with the provisions of Section 1A (Small Entities) of Financial Reporting Standard 102
+                <input class="govuk-radios__input show-text piwik-event" id="basisOfPreparation-custom-statement" data-target-text-field="basisOfPreparation-custom-statement-field" th:field="*{isPreparedInAccordanceWithStandards}" type="radio" value="0" data-event-id="FRS101">
```
BSaunders95, line 47: I don't think we want the `show-text` class at all (and the text box below the radios can be removed); the page is now going to be just the 2 radios, so there's going to be a slight change to the java logic behind the scenes

PR #511, opened 2020-08-04
Loans to directors question page

amoriartyCH:
-Create the question page for the new loans to directors work
-Unit tested.

Note, this page will change once more of the journey has been implemented.


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsQuestionController.java
```diff
@@ -0,0 +1,83 @@
...
+        setIsLoansToDirectorsIncluded(request, loansToDirectorsQuestion);
+
+        addBackPageAttributeToModel(model, companyNumber, transactionId, companyAccountsId);
+        model.addAttribute(LOANS_TO_DIRECTORS_QUESTION, loansToDirectorsQuestion);
+
+        return getTemplateName();
+    }
+
+    @PostMapping
+    public String submitLoansToDirectorsQuestion(@PathVariable String companyNumber,
```
BSaunders95, line 46: Shall we create the loans to directors parent resource within this method via a service call if one doesn't already exist?

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsQuestionController.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @GetMapping
+    public String getLoansToDirectorsQuestion(Model model,
+                                             @PathVariable String companyNumber,
+                                             @PathVariable String transactionId,
+                                             @PathVariable String companyAccountsId,
+                                             HttpServletRequest request) {
+
+        LoansToDirectorsQuestion loansToDirectorsQuestion = new LoansToDirectorsQuestion();
+        setIsLoansToDirectorsIncluded(request, loansToDirectorsQuestion);
```
BSaunders95, line 37: Since we've determined in the past that state isn't necessarily reliable, shall we fetch back the loans to directors resource here and if loans links exist, set the boolean to true, else try to derive it from the state?

src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/LoansToDirectorsServiceImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final String RESOURCE_NAME = "loans to directors";
+
+    @Override
+    public void createLoansToDirectors(String transactionId, String companyAccountsId) throws ServiceException {
+
+        ApiClient apiClient = apiClientService.getApiClient();
+
+        LoansToDirectorsApi loansToDirectors = getLoansToDirectors(apiClient, transactionId, companyAccountsId);
+
+        if (loansToDirectors == null) {
```
BSaunders95, line 42: This logic is duplicated from the controller - we already know `loansToDirectors` is null

PR #512, opened 2020-08-05
Fix a npe in the directors report for live

amoriartyCH:
-Fix null ptr when creating approval list.


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/DirectorsReportApprovalController.java
```diff
@@ -68,22 +69,19 @@ public String getDirectorsReportApproval(@PathVariable String companyNumber,
...
-                                            .collect(Collectors.toList());
+            List<String> approverOptions = new ArrayList<>();
+
+            for(Director director : directorService.getAllDirectors(transactionId, companyAccountsId)) {
+
+                if (director.getResignationDate() == null) {
+                    approverOptions.add(director.getName());
+                } else if (director.getAppointmentDate() != null) {
+                    if (director.getAppointmentDate().isAfter(director.getResignationDate())) {
+                        director.setResignationDate(null);
```
BSaunders95, line 80: This line isn't needed since `director` isn't used after this logic (besids to get the name)

PR #516, opened 2020-08-10
Remove loan functionality

amoriartyCH:
-Add the ability to remove a loan from the table.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveLoansController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveLoansController.java
@@ -139,6 +139,30 @@ public class AddOrRemoveLoansController extends BaseController implements Condit
                 URI.expand(companyNumber, transactionId, companyAccountsId).toString();
     }
 
+    @GetMapping("remove/{loanId}")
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveLoansController.java
```diff
@@ -139,6 +139,30 @@ public String addLoan(@PathVariable String companyNumber,
                 URI.expand(companyNumber, transactionId, companyAccountsId).toString();
     }
 
+    @GetMapping("remove/{loanId}")
+    public String deleteLoan(@PathVariable String companyNumber,
+                             @PathVariable String transactionId,
+                             @PathVariable String companyAccountsId,
+                             @PathVariable String loanId,
+                             @ModelAttribute(ADD_OR_REMOVE_LOANS) AddOrRemoveLoans addOrRemoveLoans,
```
BSaunders95, line 147: Is this used? I think it'd be null anyway as there's no form submission to trigger this method

amoriartyCH: well spotted, i've removed this!

PR #517, opened 2020-08-10
Implement add loan functionality

BSaunders95:
Implement add loan functionality, adding relevant input fields to the `addOrRemoveLoans` template to collect a user's loan data

Resolves BI-4413


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveLoansController.java
```diff
@@ -83,6 +115,30 @@ public String submitAddOrRemoveLoans(@PathVariable String companyNumber,
                         companyAccountsId);
     }
 
+    @PostMapping(params = "add")
```
ch-tstrickland, line 118: What is this 'add' parameter for?

amoriartyCH: When you have two buttons in a single form on thymeleaf, you can specify their (th:name="add") to let spring know which method in your controller to go to.

So add will take you to the addLoan method and submit takes you to the submit method

BSaunders95: Since there are 2 different submit buttons on the form, we need some way to differentiate between them so we've opted for request params

PR #533, opened 2020-08-25
LTD Additional Information Controller

faisal-amin82:
- Add loans to directors Additional Information selection controller and service classes
- Add html page for loans to directors selection
- Add unit tests

Resolves: BI-4474


BSaunders95:
I don't believe the generic service will do everything it should here; the logic needs to match the add / remove loans controller that if a user selects 'Yes' but the loans resource doesn't exist, we need to create the empty resource etc

amoriartyCH:
This has now been merged as part of another PR. code from this Pr was used so thanks for the help faisal!

PR #544, opened 2020-09-07
Fix issue of navigation to loans to directors additional info

mreis-ch:
Update LoansToDirectorsAdditionalController to implement ConditionalController and add willRender method
Update check in postLoansToDirectorsAdditionalInfo to check if the additional information details field is null and if so create otherwise update

BI-4475

### DIFF
--- a/src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsAdditionalInfoControllerTest.java
+++ b/src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsAdditionalInfoControllerTest.java
@@ -1,5 +1,7 @@
 package uk.gov.companieshouse.web.accounts.controller.smallfull;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.never;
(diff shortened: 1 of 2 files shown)


ch-tstrickland (commented):
Is there additional unit testing that could be done?

mreis-ch:
> Is there additional unit testing that could be done?

You're right, yes there could be. Will add some 👍

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsAdditionalInfoControllerTest.java
```diff
@@ -82,6 +88,8 @@
     private static final String MOCK_CONTROLLER_PATH = UrlBasedViewResolver.REDIRECT_URL_PREFIX + "mockControllerPath";
 
 
+    private static final String COMPANY_ACCOUNTS_DATA_STATE = "companyAccountsDataState";
```
ch-tstrickland, line 91: Please move this constant (and the others while you're at it) to the top, above the other vars.

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsAdditionalInfoControllerTest.java
```diff
@@ -172,4 +266,26 @@ void postAdditionalInformationBindingResultErrors() throws Exception {
                 .andExpect(status().isOk())
                 .andExpect(view().name(ADDITIONAL_INFORMATION_VIEW));
     }
+
+    @Test
+    @DisplayName("Will render - false")
```
ch-tstrickland, line 271: Could you add more info in this (and the following) DisplayNames? What is the condition that this is testing?

ch-tstrickland (approved):

PR #557, opened 2020-09-15
Implement LTD DR warning banner on dr page

amoriartyCH:
Implement the banner which will warn users that if they have a LTD resource with loans, and they choose to edit the DR page, then their LTD will be deleted and will need to be re-entered.

-Added unit tests where needed.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/model/directorsreport/AddOrRemoveDirectors.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/model/directorsreport/AddOrRemoveDirectors.java
@@ -6,7 +6,7 @@ import uk.gov.companieshouse.web.accounts.validation.ValidationModel;
 @ValidationModel
 public class AddOrRemoveDirectors {
 
-    private boolean displayWarningBanner;
+    private boolean displayLtdWarningBanner;
 
     private Director[] existingDirectors;
 
@@ -40,11 +40,11 @@ public class AddOrRemoveDirectors {
         this.secretary = secretary;
     }
 
-    public boolean isDisplayWarningBanner() {
-        return displayWarningBanner;
+    public boolean isDisplayLtdWarningBanner() {
+        return displayLtdWarningBanner;
     }
 
-    public void setDisplayWarningBanner(boolean displayWarningBanner) {
-        this.displayWarningBanner = displayWarningBanner;
+    public void setDisplayLtdWarningBanner(boolean displayLtdWarningBanner) {
+        this.displayLtdWarningBanner = displayLtdWarningBanner;
     }
 }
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveDirectorsControllerTest.java
```diff
@@ -127,6 +136,74 @@ void getRequestSuccess() throws Exception {
...
+                .andExpect(view().name(ADD_OR_REMOVE_DIRECTORS_VIEW))
+                .andExpect(model().attributeExists(ADD_OR_REMOVE_DIRECTORS_MODEL_ATTR))
+                .andExpect(model().attributeExists(COMPANY_NUMBER))
+                .andExpect(model().attributeExists(TRANSACTION_ID))
+                .andExpect(model().attributeExists(COMPANY_ACCOUNTS_ID))
+                .andExpect(model().attributeExists(TEMPLATE_NAME_MODEL_ATTR));
+    }
+
+    @Test
+    @DisplayName("Get add or remove directors view when a user has existing LTD loans - success path")
```
mreis-ch, line 196: How are we testing if the displayWarningBanner attribute is set to true?

src/main/java/uk/gov/companieshouse/web/accounts/model/directorsreport/AddOrRemoveDirectors.java
```diff
@@ -6,6 +6,8 @@
 @ValidationModel
 public class AddOrRemoveDirectors {
 
+    private boolean displayWarningBanner;
```
mreis-ch, line 9: Just wondering if this could be displayLtdWarningBanner as it's specifically for Loans to directors? It's a little pedantic point maybe but just a thought should any other warning banners were to be added 😄

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveDirectorsControllerTest.java
amoriartyCH: I will add an assertion at the end of each test to confirm, good point!

src/main/java/uk/gov/companieshouse/web/accounts/model/directorsreport/AddOrRemoveDirectors.java
amoriartyCH: can look to change this

mreis-ch: It's a mute point I guess. The naming suggests one banner. The likelihood is it will always be one banner just makes it more specific to LTD .

mreis-ch (approved):

PR #567, opened 2020-10-01
Sonar fixes - v1

amoriartyCH:
push 1 - more public access removal

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
@@ -89,14 +89,17 @@ public class DirectorValidator {
             error.setMessageKey(DIRECTOR_MUST_BE_ADDED);
             validationErrors.add(error);
 
-        } else if (addOrRemoveDirectors.getExistingDirectors() == null ||
-                !Arrays.stream(addOrRemoveDirectors.getExistingDirectors())
-                        .filter(d -> d.getResignationDate() == null).findFirst().isPresent()) {
+        } else {
+            assert addOrRemoveDirectors != null;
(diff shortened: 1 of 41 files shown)


mreis-ch (approved):

src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
```diff
@@ -89,14 +89,17 @@
...
-        } else if (addOrRemoveDirectors.getExistingDirectors() == null ||
-                !Arrays.stream(addOrRemoveDirectors.getExistingDirectors())
-                        .filter(d -> d.getResignationDate() == null).findFirst().isPresent()) {
-
-            ValidationError error = new ValidationError();
-            error.setFieldPath(DIRECTOR_TO_ADD);
-            error.setMessageKey(AT_LEAST_ONE_DIRECTOR_REQUIRED);
-            validationErrors.add(error);
+        } else {
+            assert addOrRemoveDirectors != null;
```
mreis-ch, line 93: Why has Sonar suggested doing this? Just want to understand

amoriartyCH: assert addOrRemoveDirectors != null; // without this addOrRemoveDir could be null and when doing .get on the object it would throw a NPE
 
It then suggested changing the .findFirst.isPresent, too an .anyMatch() call

Then finally it picked up on the NOT operator (!) and suggested changing anyMatch() too a noneMatch()

mreis-ch: I was just curious of the use of assert. If the assert fails what happens, it bypasses the block?

PR #571, opened 2020-10-01
Change assertion to not null if statement

amoriartyCH:
-Remove use of assert call and instead rely on if statement.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
@@ -103,30 +103,29 @@ public class DirectorValidator {
             }
         }
 
-        if (addOrRemoveDirectors != null) {
-            if (addOrRemoveDirectors.getExistingDirectors() != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
```diff
@@ -90,37 +90,40 @@
...
+                    error.setFieldPath(DIRECTOR_TO_ADD);
+                    error.setMessageKey(AT_LEAST_ONE_DIRECTOR_REQUIRED);
+                    validationErrors.add(error);
+                }
             }
         }
 
-        if(addOrRemoveDirectors.getExistingDirectors() != null) {
+        if (addOrRemoveDirectors != null) {
+            if (addOrRemoveDirectors.getExistingDirectors() != null) {
```
mreis-ch, line 107: Think you can combine these two if statements with &&

mreis-ch (approved):

src/main/java/uk/gov/companieshouse/web/accounts/validation/smallfull/DirectorValidator.java
amoriartyCH: As discussed, combined lines 106/107

PR #575, opened 2020-10-05
Fix Sonar issues by changing pom parent

ch-tstrickland:
Fixes failure to gather Sonar coverage reports.

We changed Sonar and, in doing so, needed to upgrade to use a more
 recent companies house parent pom. Just changing to use a newer
 companies-house-spring-boot-parent did not work breaking lots of
 tests. We are not supposed to use that as a parent any more, so
 this is a chance to use the proper companies-house-parent. In doing
 so, other changes to pull in correct spring dependencies were
 needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,8 +63,6 @@
         <maven-jar-plugin.version>2.4</maven-jar-plugin.version>
         <junit-platform-surefire-provider.version>1.2.0</junit-platform-surefire-provider.version>
 
-        <!-- Jacoco -->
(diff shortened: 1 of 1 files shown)


mreis-ch (approved):

ch-tstrickland:
Too early to approve:
https://github.com/companieshouse/company-accounts.web.ch.gov.uk/pull/575/checks?check_run_id=1209658447****
... shows that we still have the same issue.
Sorry to trouble you, but the only way to test this, as far as I can tell, is to run a PR.
I'm pushing a possible fix.

PR #586, opened 2020-11-02
Resolve issues with disappearing text

faisal-amin82:
- Resolves issue related to text `Not provided` which disappears for previous loans when validation errors fire
- Resolves issues with disappearing field values for previous loans of the same director when validation fires
- Resolves text that displays as of period start but is actually period end
- Remove unnecessary code blocks from addOrRemoveLoans.html file
- Resolves issue where if no `director` or `Prefer not to say` is not selected
- Update unit tests

Resolves: BI-5474

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/LoansServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/LoansServiceImpl.java
@@ -1,7 +1,5 @@
 package uk.gov.companieshouse.web.accounts.service.smallfull.impl;
 
-import java.util.List;
-
 import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/service/smallfull/impl/LoansServiceImpl.java
```diff
@@ -59,7 +60,14 @@
         String uri = LOANS_URI.expand(transactionId, companyAccountsId).toString();
 
         try {
-            LoanApi[] loans = apiClient.smallFull().loansToDirectors().loans().getAll(uri).execute().getData();
+            LoanApi[] loans = Arrays.stream(apiClient.smallFull().loansToDirectors().loans().getAll(uri).execute().getData())
+                    .map(loan -> {
+                        if(StringUtils.isBlank(loan.getDirectorName())) {
+                            loan.setDirectorName("Not provided");
```
amoriartyCH, line 66: Did we come to an agreement on whether "Not provided" would be used? As I remember chatting about it in a slack thread with Liam and others on Friday :)

amoriartyCH: Linked by Faisal to a previous discussion with James M, all seems okay with use of Not provided. Happy to OK pr 👍

amoriartyCH (approved):

PR #589, opened 2020-11-27
Add Related Party Transaction question page

mreis-ch:
Update api-sdk-java version to include Related Party Transaction changes
Create html for related party transaction question page
Create model, controller and service for the RelatedPartyTransactionQuestion page
Update the previous controller in OffBalanceSheetArrangementsQuestionController to the new RelatedPartyTransactionQuestionController
Update the next controller in LoansToDirectorsAdditionalInfoControler to the new RelatedPartyTransactionQuestionController
Add validation message for non-selection of radio buttons
Create unit test for page controller
Add new boolean flag with getter/setter to CompanyAccountsState
Update legend text in loansToDirectorsQuestion to include single quotes around loans to directors to make it consistent with other question pages.
Update version to 1.20

BI-5301 & BI-6087


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/RelatedPartyTransactionsQuestionController.java
```diff
@@ -0,0 +1,120 @@
...
+                                             @PathVariable String companyNumber,
+                                             @PathVariable String transactionId,
+                                             @PathVariable String companyAccountsId,
+                                             HttpServletRequest request) {
+
+        RelatedPartyTransactionsQuestion relatedPartyTransactionQuestion = new RelatedPartyTransactionsQuestion();
+
+        ApiClient apiClient = apiClientService.getApiClient();
+        try {
+            RelatedPartyTransactionsApi relatedPartyTransactionsApi = relatedPartyTransactionsService.getRelatedPartyTransactions(apiClient, transactionId, companyAccountsId);
```
amoriartyCH, line 50: Previously we inflated the _CAS cookie (company data state cookie) to a point where it was too large and was causing issues in live where the cookie would delete an "older" state to make room for a new one; this caused the WillRender method of certain notes to not display and the accounts filing had to be re-started for the user.

Could we try to use a call to the API instead? Or at least discuss whether a call to the API is better or if there could be a way to stop the cookie from getting bloated without causing potential live issues?

amoriartyCH (commented):
One small comment which is more aimed at all aardvark devs but worth putting on this PR for future reference. Other than that awesome stuff :)

src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/RelatedPartyTransactionsQuestionController.java
amoriartyCH: After a discussion with devs and Divya (lead) we have decided for now to call the API to check if a page should render or if a question has been answered. This will result in minor changes to this PR and eventually a spike/ticket to rework the _CAS cookie.

amoriartyCH: Small update, we will be relying on the _CAS cookie for now for the question page due to the fact we can't have a null selection (no selection made by user yet) if we query the API as it will only be able to return YES/NO....which would mean auto populating NO even if the user hasn't selected anything yet.

This will be investigated in a new ticket surrounding the use of the _CAS cookie.

amoriartyCH (approved):

PR #591, opened 2020-12-01
Add aria-label and role to nav element to fix axe accessibility issue

amoriartyCH:
Axe was complaining that landmarks must have a unique role, and to fix this issue:

Fix the following:
The landmark must have a unique aria-label, aria-labelledby

So I have added an aria-label and role to the footer nav element to make it unique.


ch-tstrickland (approved):
Looks good to me.

PR #592, opened 2020-12-02
Add another related-party-transaction Transaction functionality & rpt question bug fix

amoriartyCH:
Added functionality where user can now add new transactions to their existing rpt transactions. New transactions will now appear at the top of the web page in a summary table.

Unit tests have also been added to cover new code.

 bug fix regarding displaying the page logic found in the rpt question controller has also been fixed in this PR and unit test updated.

Image of new table for reference when viewing HTML of table and changes:
image


ch-tstrickland (commented):
Commit could be more detailed.
'... a bug fix': what bug fix? Hard to get an idea of what was fixed and so if it has been fixed.

ch-tstrickland (approved):

ch-tstrickland:
Sorry. Approved a little early. Fix the coverage first!

amoriartyCH:
> Commit could be more detailed.
> '... a bug fix': what bug fix? Hard to get an idea of what was fixed and so if it has been fixed.

the full commit read:
… a bug fix regarding displaying the page logic found in the rpt question controller

That i feel should describe the bug fine in a commit message

PR #594, opened 2020-12-02
Update breakdown text from loan to transaction

faisal-amin82:


mreis-ch (approved):
Might want to update PR text, misspelling of loan :)

PR #601, opened 2020-12-08
Hide balance at period start

faisal-amin82:
This commit hides the field balance at period start if accounts filer is a single year filer. For multi year filers, that field would be visible and it also deals with the added transaction and hides that field for single year filer as well

Resolves: BI-5316

Screenshot for single year filer:

image


src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveRptTransactionsControllerTest.java
```diff
@@ -16,10 +16,12 @@
 import uk.gov.companieshouse.api.ApiClient;
```
amoriartyCH, line 16: Is there a test in this ControllerTest file which checks for the MULTI_YEAR_FILER being set to false and asserting the model attribute is also false?

amoriartyCH (approved):

src/test/java/uk/gov/companieshouse/web/accounts/controller/smallfull/AddOrRemoveRptTransactionsControllerTest.java
faisal-amin82: Discussed this with Aaron and as we are mocking that functionality about setting a multi year filer, it would not be useful to assert whether the value is false or not

PR #603, opened 2020-12-08
Fix button alignment on RPT and LTD pages

mreis-ch:
An issue was found with alignment of the buttons on the Related Party Transaction page was displayed at a tablet resolution, approx 642x614. Removed the <div> around the buttons and added a new <div> above as a spacer and this has resolved the issue. Also applied the change to the Loans to Directors page as this had the same issue.

BI-5336


ch-tstrickland (approved):
Good spot on LTD too.

PR #609, opened 2021-05-06
Add a not null check to avoid a NPE on LTD

amoriartyCH:
In the loans to directors question page if a user
selects no to loans to directors, but has a ltd
additional information resource, it will attempt
to delete the loans resource but will fail with NPE
as it doesn't exist. Added a null check to avoid this.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsQuestionController.java
+++ b/src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsQuestionController.java
@@ -23,6 +23,7 @@ import uk.gov.companieshouse.web.accounts.service.smallfull.LoansToDirectorsServ
 
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
+import java.util.Map;
 
 @Controller
 @PreviousController(FinancialCommitmentsController.class)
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/LoansToDirectorsQuestionController.java
```diff
@@ -96,8 +96,10 @@ public String submitLoansToDirectorsQuestion(@PathVariable String companyNumber,
             } else {
                 if (loansToDirectorsApi != null) {
                     if (loansToDirectorsApi.getLinks().getAdditionalInformation() != null) {
-                        for (String loanId : loansToDirectorsApi.getLoans().keySet()) {
-                            loansService.deleteLoan(transactionId, companyAccountsId, loanId);
+                        if (loansToDirectorsApi.getLoans() != null) {
```
ch-tstrickland, line 99: Can the ifs on 98 & 99 be merged?

ch-tstrickland: Or convert 99-100 to:
```
LoansToDirector loanIds = loansToDirectorsApi.getLoans();
if (loanIds != null) {
    for (String loanId : loanIds) {
```

ch-tstrickland: ... which removes a call to getLoans()

amoriartyCH: Implemented option 2

ch-tstrickland (approved):
This ought to have required a unit test fix, but does not. We are planning to reduce code complexity & will address both complexity reduction & unit test fixes there.

PR #611, opened 2021-05-20
Add service map

divya051988:
- Add service map to be included in  neo4j


ch-tstrickland (approved):
This looks good... I think.
It may be good to add a property that is a link to the spec for this file format, as it is difficult to review/understand some parts. Especially integration_outgoing. In the company accounts api, we had values such as SENDS_DATA_TO, GETS_DATA_FROM, INTEGRATES_WITH...
Here we have CALLS_API.

PR #617, opened 2021-09-22
Add summary query parameter to the path to disable the summary screen

faisal-amin82:
Payment summary screen is shown by default. We do not need the summary
screen for CIC payments anymore and providing the query parameter as false,
the service goes to the payment page rather than summary screen.

Resolves: BI-8836


src/main/java/uk/gov/companieshouse/web/accounts/controller/smallfull/ApprovalController.java
```diff
@@ -44,6 +44,7 @@
     private static final String IS_PAYABLE_TRANSACTION = "isPayableTransaction";
 
     private static final UriTemplate SUBMITTED_ACCOUNTS_PATH = new UriTemplate("/company/{companyNumber}/transaction/{transactionId}/company-accounts/{companyAccountsId}/small-full/approved-accounts");
+    private static final String SUMMARY_SCREEN_FALSE = "?summary=false";
```
divya051988, line 47: keeping it consistent with other apps please could you update
SUMMARY_FALSE_PARAMETER

faisal-amin82: Changed variable name for consistency

jphillips2121 (approved):

PR #625, opened 2021-12-16
Update company-accounts-library version

faisal-amin82:
Update company-accounts-library version to have latest changes for log4j which could be tested together


CHdlewis (approved):
LGTM

PR #633, opened 2022-08-22
Use environment specific URLs

HarryBosworth:
Changed the urls in the application.properties file to use the CHS_URL environment variable so that links will link to other services in the same environment. 

Removed unused property `accounts-selector.uri`.

[BI-11618](https://companieshouse.atlassian.net/browse/BI-11618)


sbasile-ch (approved):
ok

PR #634, opened 2022-09-01
Updated error message

HarryBosworth:
Updated error message when the user doesn't click an option on the `/company/{company_number}/select-account-type` screen.

[BI-11454](https://companieshouse.atlassian.net/browse/BI-11454)


sbasile-ch (approved):
ok

PR #676, opened 2025-01-06
Add package accounts feature flag to other select-accounts-type page

HarryBosworth:
Package accounts enabled feature flag was only populated in the `/company/{companyNumber}/select-accounts-type` endpoint and not the `/select-accounts-type` endpoint.
This PR just adds the feature flag variable to the other endpoint to allow the package accounts option to be enabled.


amartin7663 (approved):

ch-code-analysis:
CI: No security warnings found

PR #677, opened 2025-01-15
Fix package accounts radio option disappearing on error

HarryBosworth:
When there is an error in the select accounts type screen the page is re-rendered with the errors, however the `packageAccountsEnabled` wasn't added back in leading to the package accounts radio button disappearing. 
This PR fixes that by adding the attribute in when the template is re-rendered with the errors. 

Resolves: [AOAF-643](https://companieshouse.atlassian.net/browse/AOAF-643)

[AOAF-643]: https://companieshouse.atlassian.net/browse/AOAF-643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ch-code-analysis:
CI: Security warnings found!

